PR #16, opened 2020-03-04
[glyf] Simplify points to contour logic

ebraminio:
If `Some(offcurve1), Some(offcurve2)` happens `last_offcurve` will be altered so on the next it won't be matched for sure and if it don't it certainly breaks leaving no need for `loop {}` AFAICS.


RazrFalcon:
Thanks!

PR #18, opened 2020-03-10
Add benchmarking against HarfBuzz draw

ebraminio:
In a static link and with -O3, which is unfair to freetype,

```
freetype_outline_glyf        1553973 ns    1551747 ns        430
freetype_outline_cff         8372932 ns    8358837 ns         85
stb_truetype_outline_glyf     900275 ns     898488 ns        763
stb_truetype_outline_cff     3625178 ns    3618126 ns        192
ttf_parser_outline_glyf       959443 ns     957739 ns        720
ttf_parser_outline_cff       1634618 ns    1632204 ns        427
harfbuzz_outline_glyf        3459615 ns    3451320 ns        204
harfbuzz_outline_cff         4129832 ns    4121364 ns        165
```

Which is disappointing but bearable for hb of course as [this](https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-draw.hh#L42-L124) is applied to our output and each point [is scaled and rounded](https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-glyf-table.hh#L1077), yet, there is room for improvements in HB I guess. Wish could just integrate ttf-parser for these anyway :/

Without returning actual numbers, thus removing scaling/rounding path begin/close check etc,

```
-----------------------------------------------------------------
Benchmark                          Time           CPU Iterations
-----------------------------------------------------------------
freetype_outline_glyf        1541238 ns    1539953 ns        455
freetype_outline_cff         8548286 ns    8539362 ns         83
stb_truetype_outline_glyf     902659 ns     901825 ns        759
stb_truetype_outline_cff     3787889 ns    3782568 ns        188
ttf_parser_outline_glyf      1014617 ns    1013474 ns        700
ttf_parser_outline_cff       1661201 ns    1659632 ns        408
harfbuzz_outline_glyf        2285818 ns    2283372 ns        303
harfbuzz_outline_cff         2881815 ns    2879962 ns        239
```

Still not good :/ Maybe I am also missing something

Couldn't setup the build however, have used harfbuzz.cc for testings which isn't good of cours.e

### DIFF
--- a/benches/outline/outline.cc
+++ b/benches/outline/outline.cc
@@ -12,6 +12,8 @@
 #define STB_TRUETYPE_IMPLEMENTATION
 #include "stb_truetype.h"
 
+#include <hb.h>
+
 namespace FT {
 struct Outliner
 {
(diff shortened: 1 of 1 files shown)


RazrFalcon:
Yes, I wanted to wait for a stable release with `draw` functions. Also, I don't want to add whole HB to the repo just for benchmarks. 
Will create issues to the meson's warpdb.

As for HB performance, yes, it's pretty bad. Not sure why, since stb_truetype also allocates, so this should not be factor. Also, ' scaling and rounding' should be essentially free. Strange to see such an impact.

PS: and I thought that ttf-parser will be the slowest one, since it's completely memory safe and very strict.

RazrFalcon:
>static link and with -O3, which is unfair to freetype

I'm using gentoo, so on my machine freetype it's built with `-march=native`. It's not -O3, which would probably brake it, but still closer to others.

Also, I'm linking ttf-parser dynamically and it doesn't affect much.

RazrFalcon:
PS: by the way, according to the FreeType [docs](https://www.freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_load_xxx), `FT_LOAD_NO_SCALE` disables hinting. So I'm not sure why FreeType is so slow.

ebraminio:
Was referring to freetype more in a sense that having a zero alloc loader
may not possible to support hinting at the same time, something I was
thinking also about gvar is proved wrong here :)

ebraminio:
> may not possible to support hinting

Should add however I understand that hinting is something belongs to past and not relevant on on increasingly adopted HiDPI screens. It is said that Apple has disabled it for much of the fonts also. As said before I like to put some effort here as has found testing and developing some new thing on Rust much easier.

Thanks for all your works :)

RazrFalcon:
The problem with hinting is that I'm not sure if it belongs to the ttf-parser side, or should it be implemented by a rasterizer. I mean, ttf-parser doesn't actually render anything. The `rusttype` (read rust's freetype) project is actually trying to use ttf-parser, but I'm not sure what is their position on hinting.

Anyway, ttf-parser can provide parsing, or even applying, of required tables, but I'm not sure if there is a demand for that. Right now I'm focused on stuff required by harfbuzz/rustybuzz.

ebraminio:
> The problem with hinting is that I'm not sure if it belongs to the ttf-parser side, or should it be implemented by a rasterizer.

Can you provide a not nesseccarily public API so one can implement it based on ttf-parser so I can play with it? Things like contour points at the right place and instrucation of the individual glyphs? Accessing fpgm and cvt is easier. The API will be useful for rusttype also I think.

RazrFalcon:
I'm not sure what api exactly do you need. Maybe it would be easier to simply fork the ttf-parser for now?

ebraminio:
Sure sure

ebraminio:
The API got hide behind a compiler flag, I think it will be important also to address its performance issues revealed here before making it public. Thanks!

RazrFalcon:
Sure. Do you have plans on releasing new HB anytime soon? The idea of backporting ~700 commits is not a pleasant one, even thought that most of them are not related to the rustybuzz.

ebraminio:
Oh, hopefully in this week.

PR #25, opened 2020-06-28
Add meson.build

ebraminio:
This enables me to use ttf-parser as a meson subproject for https://github.com/harfbuzz/harfbuzz/pull/2510 use, not sure if is the best approach, I used my fork for that for now.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,12 @@
+project('ttf-parser', 'rust')
+
+add_project_arguments(['--edition=2018'], language: 'rust')
+
+ttf_parser = static_library('ttf_parser_capi', 'c-api/lib.rs', rust_crate_type: 'staticlib',
+  link_with: static_library('ttf_parser', 'src/lib.rs'),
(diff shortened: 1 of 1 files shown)


RazrFalcon:
As long as it works for you, I'm fine.

meson.build
```diff
@@ -0,0 +1,12 @@
+project('ttf-parser', 'rust')
+
+add_project_arguments(['--edition=2018'], language: 'rust')
```
RazrFalcon, line 3: Does this necessary? Doesn't meson support Cargo.toml?

ebraminio: It doesn't officially https://github.com/mesonbuild/meson/issues/2173

ebraminio: Heated discussion here https://github.com/mesonbuild/meson/pull/2617#issuecomment-379417837 also

RazrFalcon: I see, thanks.

PR #36, opened 2020-07-21
Fix font width parse bug

### LINKED ISSUE #35: bug: ttf width is not parsed correctly
in `os2.rs#width()`, due to the default number in Rust is i32, here `Stream::read_at` will read 4 bytes instead of the length of width (which is u16, 2 bytes), leading to width never correctly parsed

zimond:
fixes #35


RazrFalcon:
Thanks!

PR #38, opened 2020-07-29
Extract more metrics from Post and OS/2 table

### LINKED ISSUE #17: Add is_monospaced method

laurmaedje:
This adds functions to extract a few more useful metrics:
- methods to get the font's `cap_height` and `italic_angle`
- an `is_monospaced` method (which fixes #17)
- three more methods to get the typo ascender, descender and line gap independently from `use_typo_metrics` (this would be useful to me because in one project I always want to use the typo metrics and not sometimes this or that depending on how the font is configured).

If you have any feedback, I'd be happy to incorporate it! I did not write any tests for this because I also didn't see tests for similarly simple table parsing stuff.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -894,10 +894,10 @@ impl<'a> Face<'a> {
 
     /// Returns face's italic angle.
     ///
-    /// Returns `0.0` when `post` table is not present.
+    /// Returns `None` when `post` table is not present.
     #[inline]
-    pub fn italic_angle(&self) -> f32 {
-        try_opt_or!(self.post, 0.0).italic_angle()
+    pub fn italic_angle(&self) -> Option<f32> {
+        self.post.map(|table| table.italic_angle())
     }
 
     #[inline]
@@ -952,7 +952,10 @@ impl<'a> Face<'a> {
         }
     }
 
-    /// Returns a horizontal typo face ascender.
+    /// Returns a horizontal typographic face ascender.
+    ///
+    /// Prefer `Face::ascender` unless you explicitly want this. This is a more
+    /// low-level alternative.
     ///
     /// This method is affected by variation axes.
     ///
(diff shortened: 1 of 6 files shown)


RazrFalcon:
Would you mind updating the C API too? It should be trivial.

PS: yeah, tests are in progress...

src/lib.rs
```diff
@@ -884,6 +892,14 @@ impl<'a> Face<'a> {
         try_opt_or!(self.os_2, Width::default()).width()
     }
 
+    /// Returns face's italic angle.
+    ///
+    /// Returns `0.0` when `post` table is not present.
+    #[inline]
+    pub fn italic_angle(&self) -> f32 {
```
RazrFalcon, line 899: I think it should return `Option<f32>`. And just float/0.0 in C API.

src/lib.rs
```diff
@@ -936,6 +952,45 @@ impl<'a> Face<'a> {
         }
     }
 
+    /// Returns a horizontal typo face ascender.
```
RazrFalcon, line 955: Could you note in the docs, that a user should prefer ascender/descender methods. And those one are more low level.

Also, could you use *typographic* instead of *typo*?

src/lib.rs
```diff
@@ -992,6 +1047,17 @@ impl<'a> Face<'a> {
             .map(|v| self.apply_metrics_variation(Tag::from_bytes(b"xhgt"), v))
     }
 
+    /// Returns face's cap height.
+    ///
+    /// This method is affected by variation axes.
+    ///
+    /// Returns `None` when OS/2 table is not present or when its version is < 2.
+    #[inline]
+    pub fn cap_height(&self) -> Option<i16> {
```
RazrFalcon, line 1056: Could you rename it into `capital_height`? And in the doc comment too.

laurmaedje:
I incorporated your feedback :)

RazrFalcon:
Nice, but you haven't changed the typo_* methods names.

laurmaedje:
Changed it, I thought you were only referring to the doc-comments.

RazrFalcon:
You haven't updated the `ttfparser.h` (you should do it manually, not via cbindged) and the changelog has old method names.

Other than that - it's perfect. Thanks for the contribution. I will publish a new release after the merge.

PR #39, opened 2020-07-30
Codepoints (fixes #20)

### LINKED ISSUE #20: support reading all available code-points out of the cmap table.
I have an implementation of this in https://github.com/JAicewizard/ttf-parser, I have not been able to test this for all formats but for the font's I've been able to test with it has worked.

Should I make a pull-request? The reading style of the getting a glyph id out of the cmap is something I am not used to and decided not to use for my own sake, I understand it makes it more difficult to accept.

laurmaedje:
I implemented reading all codepoints from a subtable through `Subtable::codepoints` as laid out in #20.

As discussed, this only parses the codepoints and not the glyph indices. Normally, instead of returning glyph ID `0`, the parsing methods return `None`. Since the codepoint-methods don't know about the glyph indices, calling `glyph_index` on a collected codepoint may now still return `None`. But I think that isn't too bad since a user can still check whether `glyph_index` returns `None` for each collected codepoint if necessary.

There is one exception to this rule for format 0 because otherwise we would always return all codepoints from `0` to `255` even though some of them may definitely be unmapped (aka mapped to zero).

While implementing format 0 and format 2 I noticed two things which I believe are bugs. I took the liberty to fix them in this PR:
- For format 0, the parser would return glyph ID `0` instead of `None` even though (at least I think) zero means unmapped for this table (as it has fixed size and otherwise couldn't support fonts with less than 256 mappings)
- For format 2, there was a `>` comparison with an exclusive range end which should have been `>=` as far as I can tell. This could lead to a codepoint being mapped to a glyph when it was in fact unmapped.

I added tests for the two bugs and for the `codepoints` for formats 0 and the more complex ones 2 & 4.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Added
+- `Subtable::codepoints`
 
 ## [0.8.1] - 2020-07-29
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
-- `Subtable::codepoints`
+- `cmap::Subtable::codepoints`
 
 ## [0.8.1] - 2020-07-29
 ### Added
--- a/src/tables/cmap/format0.rs
+++ b/src/tables/cmap/format0.rs
@@ -15,3 +15,44 @@ pub fn parse(data: &[u8], code_point: u32) -> Option<u16> {
         None
     }
 }
+
+pub fn codepoints(data: &[u8], mut f: impl FnMut(u32)) {
+    let mut s = Stream::new(data);
+    s.skip::<u16>(); // format
+    s.skip::<u16>(); // length
+    s.skip::<u16>(); // language
+
+    for code_point in 0..256 {
+        // In contrast to every other format, here we take a look at the glyph
+        // id and check whether it is zero because otherwise this method would
+        // always simply call `f` for `0..256` which would be kind of pointless
+        // (this array always has length 256 even when the face has only fewer
+        // glyphs).
+        let glyph_id: u8 = try_opt_or!(s.read(), ());
+        if glyph_id != 0 {
+            f(code_point);
+        }
+    }
+}
+
+#[cfg(test)]
+mod format0_tests {
+    use super::codepoints;
+
+    #[test]
+    fn maps_not_all_256_codepoints() {
+        let mut data = vec![
+            0x00, 0x00, // format: 0
+            0x01, 0x06, // subtable size: 262
+            0x00, 0x00, // language ID: 0
+        ];
+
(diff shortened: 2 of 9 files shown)


RazrFalcon:
I think you should resend the subtable 0 and 2 fixes as separated pull requires. So we can merge them before this one.

As for the patch itself, I think you should prefer using an inner method instead of constant `try_opt_or`. Basically:

```rust
pub fn codepoints(data: &[u8], mut f: impl FnMut(u32)) {
    let _ = codepoints_impl(data, f);
}

fn codepoints_impl(data: &[u8], mut f: impl FnMut(u32)) -> Option<()> {
    // you can use ? here
    Some(())
}
```

src/tables/cmap/mod.rs
```diff
@@ -180,6 +180,54 @@ impl<'a> Subtable<'a> {
...
+    /// Note that this may list codepoints for which `glyph_index` still returns
+    /// `None` because this method finds all codepoints which were _defined_ in
+    /// this subtable. The subtable may still map them to glyph ID `0`.
+    ///
+    /// Returns without doing anything:
+    /// - when format is `MixedCoverage`, since it's not supported.
+    /// - when format is `UnicodeVariationSequences`, since it's not supported.
+    pub fn codepoints<F>(&self, f: F)
+    where
+        F: FnMut(u32),
```
RazrFalcon, line 199: I think this can be inlined, e.g. `pub fn codepoints<F: FnMut(u32)>(&self, f: F)`

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Added
+- `Subtable::codepoints`
```
RazrFalcon, line 9: It should be `cmap::Subtable::codepoints`, because we also have `kern::Subtable`.

laurmaedje:
I changed the three requested things. The two bug fixes are now separate PRs. Once these are merged there will probably be merge conflicts here because of the tests and changelog.

RazrFalcon:
I've merged other patches.

src/tables/cmap/format0.rs
```diff
@@ -15,3 +15,46 @@ pub fn parse(data: &[u8], code_point: u32) -> Option<u16> {
...
+        if glyph_id != 0 {
+            f(code_point);
+        }
+    }
+
+    Some(())
+}
+
+#[cfg(test)]
+mod format0_tests {
```
RazrFalcon, line 41: `tests` are enough. No need to use the `formatN` prefix.

laurmaedje: Okay, I did that because it was like that for format 4 but I can change it.

src/tables/cmap/format4.rs
```diff
@@ -66,9 +66,33 @@ pub fn parse(data: &[u8], code_point: u32) -> Option<u16> {
...
+        }
+    }
+
+    Some(())
+}
+
 #[cfg(test)]
 mod format4_tests {
-    use super::parse;
+    use super::{parse, codepoints};
```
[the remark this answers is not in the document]
laurmaedje: Sure

RazrFalcon:
Thanks again for the contributions!

PR #47, opened 2021-01-15
Fix edge case in seac_code_to_glyph_id and clarify behaviour

wezm:
Saw the comment about 228/zcaron and set out to find an explanation. The reason is that the ISO Adobe charset only defines string ids up to 228 (p. 47 of [CFF Spec](https://wwwimages2.adobe.com/content/dam/acom/en/devnet/font/pdfs/5176.CFF.pdf)). Then the `<` stood out to me as it seems like 228 is valid too. Confirmed this by checking Harfbuzz:

https://github.com/harfbuzz/harfbuzz/blob/c8c5e52aba904cfede1c7e8e1043a392053938e6/src/hb-ot-cff1-table.hh#L1166


RazrFalcon:
Thanks!

PR #51, opened 2021-01-21
Exclude fonts from package

rtsuk:
I'm using this crate in Fuchsia, where we vendor all third party crates. Our policy is to not host font files, so it would be great if you could accept this change that would prevent them from being in the package and thus vendored.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/RazrFalcon/ttf-parser"
 documentation = "https://docs.rs/ttf-parser/"
 readme = "README.md"
 edition = "2018"
-exclude = ["benches/**"]
+exclude = ["benches/**", "fonts/**"]
 
 [features]
 default = ["std", "variable-fonts"]


RazrFalcon:
This font used here: https://github.com/RazrFalcon/ttf-parser/blob/50502770a397b0010431adca79e03a18cb839f1e/src/lib.rs#L1486 So tests would fail in case of a crate package.

I'm not sure about the policy not to store fonts, because I plan to use handcrafted fonts in the future for testing. Does it violates your rules too?

rtsuk:
If the licensing of the font is the same as the source code for the crate it will be fine.

RazrFalcon:
Then I will create a dummy font with a single glyph and replace this one. The `outline` function is pretty critical and I want to keep a working example.

rtsuk:
Thanks!

RazrFalcon:
Done. Should I publish a new release?

rtsuk:
Yes please.

PR #52, opened 2021-02-03
Split out font table parsing into separate struct

fschutt:
This enables support for custom font table providers, which is necessary if you want to decode the font tables in some way before handing them to ttf-parser (i.e.: gzip decoding / brotli decoding, etc.). 

Due to the `Deref` trait impls, this changes nothing about the public API - all methods that were previously available on the `Font` struct are now available on the `ParsedFaceTables` struct, too.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -39,6 +39,7 @@ extern crate std;
 use core::fmt;
 use core::num::NonZeroU16;
 use core::ops::{Deref, DerefMut};
+use crate::parser::LazyArrayIter16;
 
 macro_rules! try_opt_or {
     ($value:expr, $ret:expr) => {
@@ -697,30 +698,6 @@ impl<'a> Face<'a> {
         s.advance(6); // searchRange (u16) + entrySelector (u16) + rangeShift (u16)
         let tables = s.read_array16::<TableRecord>(num_tables)
             .ok_or(FaceParsingError::MalformedFont)?;
-
-        use crate::parser::LazyArrayIter16;
-
-        struct DefaultTableProvider<'a> {
-            data: &'a [u8],
-            tables: LazyArrayIter16<'a, TableRecord>,
-        }
-
-        impl<'a> Iterator for DefaultTableProvider<'a> {
-            type Item = Result<(Tag, Option<&'a [u8]>), FaceParsingError>;
-
-            // next() is the only required method
-            fn next(&mut self) -> Option<Self::Item> {
-                self.tables.next().map(|table| {
-                    Ok((table.table_tag, {
-                        let offset = usize::num_from(table.offset);
-                        let length = usize::num_from(table.length);
-                        let end = offset.checked_add(length).ok_or(FaceParsingError::MalformedFont)?;
-                        let range = offset..end;
-                        self.data.get(range)
-                    }))
-                })
-            }
-        }
 
         let internal = FaceTables::from_table_provider(
             DefaultTableProvider {
@@ -1778,6 +1755,28 @@ impl<'a> FaceTables<'a> {
     #[inline]
     fn coords(&self) -> &[NormalizedCoordinate] {
         self.coordinates.as_slice()
+    }
+}
+
+struct DefaultTableProvider<'a> {
+    data: &'a [u8],
+    tables: LazyArrayIter16<'a, TableRecord>,
+}
+
+impl<'a> Iterator for DefaultTableProvider<'a> {
+    type Item = Result<(Tag, Option<&'a [u8]>), FaceParsingError>;
+
+    // next() is the only required method
+    fn next(&mut self) -> Option<Self::Item> {
+        self.tables.next().map(|table| {
+            Ok((table.table_tag, {
+                let offset = usize::num_from(table.offset);
+                let length = usize::num_from(table.length);
+                let end = offset.checked_add(length).ok_or(FaceParsingError::MalformedFont)?;
+                let range = offset..end;
+                self.data.get(range)
+            }))
+        })
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -590,18 +590,20 @@ pub struct Face<'a> {
 impl<'a> Deref for Face<'a> {
     type Target = FaceTables<'a>;
 
+    #[inline]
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -583,6 +584,32 @@ impl std::error::Error for FaceParsingError {}
...
+
+/// Parsed face tables that are stored separately from the main font
+///
+/// This struct adds the `from_table_provider()` method that is not
+/// available on the `Font` - you can create a `ParsedFaceTables` struct
+/// from your own, custom font provider - this is important if your font
+/// provider does things that ttf-parser currently doesn't implement
+/// (for example zlib / brotli decoding)
+#[derive(Clone)]
+pub struct ParsedFaceTables<'a> {
```
RazrFalcon, line 612: Rename to just `FaceTables`.

src/lib.rs
```diff
@@ -583,6 +584,32 @@ impl std::error::Error for FaceParsingError {}
...
+    }
+}
+
+impl<'a> DerefMut for Face<'a> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.internal
+    }
+}
+
+/// Parsed face tables that are stored separately from the main font
```
RazrFalcon, line 604: Missing dot.

_Parsed face tables._ is enough.

src/lib.rs
```diff
@@ -583,6 +584,32 @@ impl std::error::Error for FaceParsingError {}
...
+///
+/// This struct adds the `from_table_provider()` method that is not
```
RazrFalcon, line 606: Could you use `.` instead of ` - `?

src/lib.rs
```diff
@@ -583,6 +584,32 @@ impl std::error::Error for FaceParsingError {}
 pub struct Face<'a> {
     font_data: &'a [u8], // The input data. Used by Face::table_data.
     table_records: LazyArray16<'a, TableRecord>,
+    internal: ParsedFaceTables<'a>,
+}
+
+impl<'a> Deref for Face<'a> {
+    type Target = ParsedFaceTables<'a>;
+
+    fn deref(&self) -> &Self::Target {
```
RazrFalcon, line 593: Add `#[inline]`

src/lib.rs
```diff
@@ -583,6 +584,32 @@ impl std::error::Error for FaceParsingError {}
...
+        &self.internal
+    }
+}
+
+impl<'a> DerefMut for Face<'a> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
```
RazrFalcon, line 599: Add `#[inline]`

src/lib.rs
```diff
@@ -663,9 +697,66 @@ impl<'a> Face<'a> {
         let tables = s.read_array16::<TableRecord>(num_tables)
             .ok_or(FaceParsingError::MalformedFont)?;
 
-        let mut face = Face {
+        use crate::parser::LazyArrayIter16;
+
+        struct DefaultTableProvider<'a> {
```
RazrFalcon, line 702: Extract to the module itself.

src/lib.rs
```diff
@@ -663,9 +697,66 @@ impl<'a> Face<'a> {
...
+        let (_, table) = self.table_records.binary_search_by(|record| record.table_tag.cmp(&tag))?;
+        let offset = usize::num_from(table.offset);
+        let length = usize::num_from(table.length);
+        let end = offset.checked_add(length)?;
+        self.font_data.get(offset..end)
+    }
+}
+
+impl<'a> ParsedFaceTables<'a> {
+
```
RazrFalcon, line 751: Remove empty line.

src/lib.rs
```diff
@@ -663,9 +697,66 @@ impl<'a> Face<'a> {
...
+    /// Creates and parses face tables from an existing table provider
```
RazrFalcon, line 752: Missing dot.

src/lib.rs
```diff
@@ -785,7 +874,6 @@ impl<'a> Face<'a> {
 
         Ok(face)
     }
-
     /// Checks that face has a specified table.
```
RazrFalcon, line 877: Bring back this empty line.

src/lib.rs
```diff
@@ -583,6 +584,32 @@ impl std::error::Error for FaceParsingError {}
...
+
+/// Parsed face tables that are stored separately from the main font
+///
+/// This struct adds the `from_table_provider()` method that is not
+/// available on the `Font` - you can create a `ParsedFaceTables` struct
+/// from your own, custom font provider - this is important if your font
+/// provider does things that ttf-parser currently doesn't implement
+/// (for example zlib / brotli decoding)
+#[derive(Clone)]
+pub struct ParsedFaceTables<'a> {
```
RazrFalcon, line 613: Remove empty line.

RazrFalcon (changes requested):

RazrFalcon (approved):

RazrFalcon:
Everything is fine now. 

Honestly, I don't really like this change. It complicates everything even more. But I must admit that yours solution is probably the best one.

PS: do I understand correctly, that WOFF compresses each table separately? Or why do you want to parse tables in this way?

fschutt:
I want to initialize the tables from decoded WOFF tables. In the WOFF2 file format, font tables are brotli-encoded as a single block. `ttf-parser` does not support brotli decoding (and that's okay, because I already use allsorts to decode the actual font). However, now I want to get the glyph outlines (using ttf-parser) - so I need some way to pass the already-decoded font tables to ttf-parser, so that I can call `get_outline` on the `FontTables`.

I know it's not great, but at first I tried copying the outline-parsing code into allsorts, but that was an even worse solution, so I left it at that. Right now I just let allsorts unzip the font tables, and then I create a cache of all outlines in the font (memory expensive, I know).

RazrFalcon:
Can't you decode WOFF first and then pass it to allsorts/ttf-parser?

RazrFalcon:
I've looked at the WOFF spec and looks like it compresses tables separately, therefore your patch is necessary anyway.

fschutt:
okay, could you release a new version so I can depend on it? thanks

RazrFalcon:
Done.

PR #58, opened 2021-05-24
0.12.1: prepare extra derives

dhardy:
This is more than what I require, but I think all make sense *unless* there is a chance they may have incompatible fields such as `f32` in the future.

Since the PR to `fontdb` will depend on this and it's non-breaking, I marked this as a release (didn't check whether anything else needs adding to the changelog).


RazrFalcon:
Good. But could you remove the last commit?

dhardy:
Done.

PR #67, opened 2021-10-19
Change ascender/descender logic (fixes #66)

### LINKED ISSUE #66: Should typographic ascender/descender be default for fonts with OS/2 table version < 4?
Hello,

A recent of upgrade of the ttf-parser dependency in the femtovg crate has caused an issue with alignment of the icons (provided  by characters in the entypo font), as can be seen here: <https://github.com/femtovg/femtovg/issues/56>.

As noted in the issue, this problem has been tracked down to https://github.com/RazrFalcon/ttf-parser/commit/d32a5dee339905976a858013a2a5a66354a90309, which added windows acender/ descender if the USE_TYPO_METRICS flag is not set.

Looking into this I think the problem may be in the `is_use_typo_metrics()` function here: <https://github.com/RazrFalcon/ttf-parser/blob/9ab844b496f21140a5e9f223bbbb1f5f89510ac3/src/tables/os2.rs#L268-L274>

This function returns false is the OS/2 table version is less than 4, which makes sense because the USE_TYPO_METRICS  flag was only introduced in version 4, but this means that the above function will cause the selection of ascender/descender to default to windows. The entypo font has an OS/2 table which is version 3. Now I can't find anything in the docs to suggest what the default ascender/descender selection should be for fonts with a version less than 4, but since it seems to be broken I would guess that maybe it should default to typographic.

My knowledge of font tables is limited so let me know if I'm way off here and if the solution should be on the femtovg usage side of things. Thanks.

geom3trik:
Applied the logic found here, <https://github.com/freetype/freetype/blob/49270c17011491227ec7bd3fb73ede4f674aa065/src/sfnt/sfobjs.c#L1306-L1311>, to ttf-parser ascender/descender calculation. Not entirely sure if the same logic should have been applied to the line gap.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1042,13 +1042,28 @@ impl<'a> FaceTables<'a> {
     /// This method is affected by variation axes.
     #[inline]
     pub fn ascender(&self) -> i16 {
+        // 1. If the OS/2 table exists and the fsSelection bit 7 is set
+        // (USE_TYPO_METRICS), trust the font and use the sTypo* metrics.
+        // 2. Otherwise, use the `hhea' table's metrics.
+        // 3. If they are zero and the OS/2 table exists,
+        //    1. use the OS/2 table's sTypo* metrics if they are non-zero.
+        //    2. Otherwise, use the OS/2 table's usWin* metrics.
         if let Some(os_2) = self.os_2 {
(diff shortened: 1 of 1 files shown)


geom3trik:
Not sure if this is linking correctly but this is a potential fix for #66

PR #94, opened 2022-04-23
bump dev-dependency base64 from version 0.12 to 0.13

alexanderkjall:


RazrFalcon:
Thanks, but it's fine as is.

PR #97, opened 2022-05-10
docs: add COLR/CPAL table and update FreeType information

yisibl:
- Add `COLR/CPAL` tables
- FreeType 2.10.0 support `COLR/CPAL` tables
- FreeType 2.11.0 support `COLR` table v1 ([COLRv1](https://caniuse.com/colr-v1))
- [FreeType 2.12.0](https://github.com/freetype/freetype/commit/21d0fa374200aecc6605daf034a47167ea82ed6f) support `SVG` table(OT-SVG)
  > FreeType now handles OT-SVG fonts, to be controlled with
FT_CONFIG_OPTION_SVG configuration macro. By default, it can
only load the 'SVG ' table of an OpenType font. However, by using
the svg-hooks property of the new 'ot-svg' module it is possible
to register an external SVG rendering engine. The FreeType demo
programs have been set up to use 'librsvg' as the rendering
library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,9 @@ There are roughly three types of TrueType tables:
 | `bloc` table      |                        | ✓                   |                                |
 | `CBDT` table      | ✓                      | ✓                   |                                |
 | `CBLC` table      | ✓                      | ✓                   |                                |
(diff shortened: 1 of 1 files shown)


yisibl:
Thanks

PR #100, opened 2022-06-29
make fields of lookup subtables public

tizee:
A small fix on lookup subtables as the visibility of fields of a public struct are default to private, which means they cannot be accessed from the outside of the module.

```rust
    if let Some(table) = face.tables().gsub {
        for lookup in table.lookups {
            for subtable in lookup.subtables {
                // process subtable
                println!("{}", subtable.kind);
            }
     }
 }
``` 

Use case:
- Some applications may need to manipulate these lookup tables manually so I need access to its fields. (Recently I was developing an text rendering application based on `ttf-parser`'s `LookupSubtables`)

BTW, this project does not have a convention of code style .i.e a `rustfmt` file. So if code style matters, I am willing to revert these unrelated changes (The formatting is performed by `rustfmt` with its default config).

### DIFF
--- a/src/ggg/lookup.rs
+++ b/src/ggg/lookup.rs
@@ -46,9 +46,12 @@ pub trait LookupSubtable<'a>: Sized {
 /// A list of lookup subtables.
 #[derive(Clone, Copy)]
 pub struct LookupSubtables<'a> {
-    kind: u16,
-    data: &'a [u8],
-    offsets: LazyArray16<'a, Offset16>,
+    /// lookup type for GPOS or GSUB
+    pub kind: u16,
+    /// raw data of subtables
+    pub data: &'a [u8],
+    /// offsets to each lookup subtables
+    pub offsets: LazyArray16<'a, Offset16>,
 }
 
 impl core::fmt::Debug for LookupSubtables<'_> {


tizee:
> I personally don't like code formatters and therefore don't use them. At the same time I'm basically following the default rustfmt logic with some small exceptions. So for a new file you could use auto-formatting, but leave other files as is.

Now I see that.

tizee:
I'd like t filter out all specific subtables via `kind` before parsing them.

I have to iterate over all subtables to parse them into corresponding types.
```rust
            let tables: Vec<_> = lookup
                .subtables
                .into_iter()
                .collect();
```
Then I could filter ones I need with patterns. 

As the code is intentionally designed for hiding all internal data, there is no need for using `kind`. Instead of that, we should use corresponding types with patterns for this.

Use `LookupSubtables::get(offset)` to parse a subtable is not-straightforward.

```
    pub fn get<T: LookupSubtable<'a>>(&self, index: u16) -> Option<T> {
        let offset = self.offsets.get(index)?.to_usize();
        let data = self.data.get(offset..)?;
        T::parse(data, self.kind)
}
```
The indices of these subtables are unknown beforehand so it discourages the use of this function. But it's not the concern of this PR.

Thanks for answering. I'll close this PR.

RazrFalcon:
So do you found a solution or still locking for it? I know that OpenType tables API is very complex, but this is because those table are complex. 

The reason `kind` is also hidden, is because it's value depends on which table you're parsing. GPOS and GSUB has their own types.

Are sure that parsing a GPOS/GSUB subtable is that expensive?

I this is absolutely needed, I can add `LookupSubtables::get_kind` that will return a corresponding enum.
As the comment of the `get()` method mentions, `LookupSubtables` should be generic by itself. But it would require a significant rewrite. Will see.

https://github.com/RazrFalcon/ttf-parser/issues/101

PR #106, opened 2022-09-30
[Fixes #105] Treat regular font as a one element collection when parsing

### LINKED ISSUE #105: When parsing a regular font (not a collection) the index is ignored
It should return `FaceParsingError::FaceIndexOutOfBounds` when reading from an index different from 0 instead.

An example code that represents what I'm talking about:
```rust
let data = FONT;

assert_eq!(fonts_in_collection(&data), None);
assert_eq!(
    Face::parse(&data, 1).unwrap_err(),
    FaceParsingError::FaceIndexOutOfBounds
);
```

Right now the second assertion fails.

Pietrek14:
When parsing a font that is not a collection, a check is performed that returns a `FaceParsingError::FaceIndexOutOfBounds` error if the index is not zero. It's a breaking change, but it's not hard to migrate to the new version.


RazrFalcon:
Thanks. The test should be in `tests/tables/main.rs`. And there is no need to include the font file itself as well. The header part should be enough. Like this:
```rust
#[test]
fn font_index_overflow_on_regular_font() {
    use Unit::*;
    let data = convert(&[
        Raw(&[0x00, 0x01, 0x00, 0x00]), // magic
        UInt16(0),                      // numTables
        UInt16(0),                      // searchRange
        UInt16(0),                      // entrySelector
        UInt16(0),                      // rangeShift
    ]);

    assert_eq!(fonts_in_collection(&data), None);
    assert_eq!(
        Face::parse(&data, 1).unwrap_err(),
        FaceParsingError::FaceIndexOutOfBounds
    );
}
```

RazrFalcon:
Thanks!

PR #107, opened 2022-11-01
ignore-singularity-contours

Shubhamj280:
Some fonts have single point contours which cause weird behaviour when stroked.
This MR ignores such contours which only have single points.

### DIFF
--- a/src/tables/glyf.rs
+++ b/src/tables/glyf.rs
@@ -78,6 +78,9 @@ impl<'a> Builder<'a> {
     #[inline]
     pub fn push_point(&mut self, x: f32, y: f32, on_curve_point: bool, last_point: bool) {
         let p = Point { x, y };
+        if self.first_on_curve.is_none() && last_point {
+            return; // Ignore "singularity contours"
+        }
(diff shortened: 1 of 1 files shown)


RazrFalcon:
I'm sorry, but I don't think this is the right solution. And I'm not even sure we should skip such contours to begin with. This should be handled by the caller.

Shubhamj280:
> I'm not even sure we should skip such contours to begin with. This should be handled by the caller.

Im a little conflicted about this point myself, but there isnt any information in the Outline builder about how we can handle these cases particularly on caller side.

> I'm sorry, but I don't think this is the right solution.

Im not sure I understand, why wouldnt this work. `first_on_curve` being `none` would imply there were no more points before this point, and `is_last_point` would mean this is also the last point.

RazrFalcon:
Can you check what other libraries are doing in this case? The `benches/outline/` has code sample for stb_truetype and freetype2. If they produce the same output for your font as ttf-parser then this is definitely should be handled by the caller.

RazrFalcon:
Out of scope.

PR #108, opened 2022-11-01
return glyph ids along with svg image

Shubhamj280:
A single svg document can be used to represent multiple glyphs. So added function 'glyph_svg_image_and_glyph_ids' to return both the svg document and the glyph ids represented by that document.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1673,6 +1673,15 @@ impl<'a> Face<'a> {
         self.tables.svg.and_then(|svg| svg.documents.find(glyph_id))
     }
 
+    /// Returns a reference to a glyph's SVG image and a range of glyph ids that are
+    /// represented using the same svg document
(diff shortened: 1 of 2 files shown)


RazrFalcon:
I see the point, but it should be implemented differently. You could simply make `SvgDocumentRecord` partially public. `svg_doc_offset` and `svg_doc_length` should still be private, but everything else can be declared public.

Shubhamj280:
Im sorry, I dont exactly understand how that will work.
`SvgDocumentsList.get` function uses the `SvgDocumentRecord `internally but it is never returned, so how will making its range based member public help. 
If you could elaborate on your thought, that would help. Thanks!

wjian23:
I encountered the same problem, please check the mr: https://github.com/RazrFalcon/ttf-parser/pull/132

Issue #110, opened 2023-01-09
Zero offsets in `parse_at_offset16`


laurmaedje:
The `parse_at_offset16` method currently does not check for NULL offsets, leading to bugs where 0 indicates the absence of a subtable. (I experienced the bug [in the math table][bug].) Not all OpenType offsets allow 0, but I don't think 0 is ever a valid offset, so I assume it should be fine to just always return `None` for zero offsets in `parse_at_offset16`. If you agree, I can send a PR with fix.

[bug]: https://github.com/RazrFalcon/ttf-parser/blob/a38c74fd1bb8f9daa24ba673791d35ff1d5c972d/src/tables/math.rs#L818

RazrFalcon:
>so I assume it should be fine to just always return None for zero offsets

I would not assume anything in case of TrueType.

As you correctly noted, not all offsets are allowed to be NULL and for now you have to check them manually.
I would suggest adding `parse_at_non_null_offset16` and use it only where the spec allows NULL. Because technically, a NULL value set for an offset that doesn't allow NULL is a hard error and we have to skip the whole table.

In other parts of the codebase I specifically use `s.read::<Option<Offset16>>()` for this. Ugly, but this is how it is.

laurmaedje:
From the spec:
> A NULL subtable offset always indicates that the given subtable is not present. Applications should never interpret a NULL offset value as the offset to subtable data. For cases in which subtable offset fields are not documented as permitting NULL values, font compilers must include a subtable of the indicated format, even if it is a header stub without further data (for example, a [coverage](https://learn.microsoft.com/en-us/typography/opentype/spec/chapter2#coverage-table) table with no glyph IDs). Applications parsing font data should, however, anticipate non-conformant font data that has a NULL subtable offset where only a non-NULL value is expected.

Adding a new `parse_at_non_null_offset16` and keeping the current `parse_at_offset16` as-is conflicts with "applications should never interpret a NULL offset value as the offset to subtable data", doesn't it? And if we adapt `parse_at_offset16` to conform with this, what would be the difference between the two methods?

RazrFalcon:
Can you give a link to this doc? I'm not sure if it's talking about the specific table or all of them.

Ok, my bad. I got confused with `parse_at_offset16` and `read_at_offset16`. The first one isn't used anywhere except for MATH. But I guess you're proposing to change `read_at_offset16` as well?
This would be a breaking change with unknown consequences.

laurmaedje:
[...]

PR #111, opened 2023-01-09
Handle null offsets in MATH table

laurmaedje:
This lets math subtables be `None` instead of attempting to read at an offset of `0`. As discussed in #110. It also moves the `read_at_offset32` to a private stream extension for the `ankr` table as discussed. While this doesn't fully resolve the offset problem, it at least fixes the specific bug in the math table.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -775,19 +775,6 @@ impl<'a> Stream<'a> {
         let offset = self.read::<Offset16>()?.to_usize();
         data.get(offset..)
     }
-
-    #[allow(dead_code)]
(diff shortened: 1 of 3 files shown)


RazrFalcon:
Ok, good for now. Do you need a version bump?

I might look for a better solution on the weekends.

laurmaedje:
Version bump would be nice, but I can also wait for after the weekend. :)

RazrFalcon:
Done.

laurmaedje:
Thanks!

PR #113, opened 2023-02-12
Replace deprecated core::i32::MIN/MAX with standard ones.

Skaldebane:
`core::i32::MIN` and `core::i32::MAX` have been deprecated, and this commit replaces them with `i32::MIN` and `i32::MAX` respectively.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.42.0
+          - 1.43.0
           - stable
     steps:
     - name: Checkout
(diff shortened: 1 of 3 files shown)


RazrFalcon:
Seems like it need 1.43
You can bump MSRV as well.

Skaldebane:
Bumped MSRV to 1.43.0

RazrFalcon:
Weird, looks like Cargo is broken. Will rerun later.

RazrFalcon:
Let's leave it for later.

Skaldebane:
Mmm, at the time of making the pull request, it was an IDE suggestion and it was compiling successfully.

However trying to use i32::MAX right now (Rust 1.70) shows that `Constant 'MAX' is private [E0603]`, indicating that it's private for some reason? It's a very useful constant so I'm not sure why they're deprecating the old way with no replacement (since the new method isn't really public API).

This also applies to all signed types (from i8 to i128). `MAX` isn't private on unsigned integers though.

RazrFalcon:
Yeah, not great. Better leave it as is as long as it compiles.

PR #115, opened 2023-02-13
Add missing loca table to FaceTables

Skaldebane:
The `loca` table is missing from `FaceTables` for some reason, this PR adds it.


RazrFalcon:
This is intentional. We don't have `CBLC` either.

`FaceTables` is a somewhat high-level API.

Skaldebane:
Ah alright, understood 👌🏻

PR #118, opened 2023-03-08
Ignoring "singularity contours" - contours that have nothing other than a single move

nwoods-cimpress:


RazrFalcon:
Yes, this is a know "issue". The problem is that I'm not sure how zero-length subpaths should be handled. Afaik this is undocumented. For now, the correct way is to handle it on the caller side.

PR #121, opened 2023-04-12
Support for bitmap fonts, EBDT + EBLC tables

dzamkov:
This PR makes a few changes that allow `ttf_parser` to be used with bitmap fonts (i.e. `.otb` files):
* Added EBDT and EBLC tables (which are basically just aliases for CBDT and CBLC)
* Implemented additional formats for the CBLC parser
* Extended `RasterImageFormat` with the various uncompressed bitmap formats that may appear in a font

The PR also fixes some issues with CBLC parsing I found during testing:
* The code for searching through subtables to find the best size did not correctly track the byte stream. There are *3* bytes between the end of `ppem_x` and the start of the next subtable, not 4. Also, these final bytes always have to be skipped, not just when `!(start_glyph_id..=end_glyph_id).contains(&glyph_id)`.
* `location.metrics` was never populated and always held the default value

I tested the changes on [Terminus](https://packages.debian.org/sid/fonts-terminus-otb) fonts and on some fonts I exported using [Bits'N'Picas](https://github.com/kreativekorp/bitsnpicas). The Terminus fonts are monochrome with several subtables for different scales. The Bits'N'Picas fonts are 8-bit grayscale.

The PR includes an automated test using one limited example font I created myself (so there are no copyright issues).


RazrFalcon:
Hi! Thanks for the patch. There are a couple of reasons why EBDT+EBLC haven't been implemented yet.

1. I haven't been able to find a font that uses those tables. Even CBDT+CBLC are exceedingly rare. Afaik only Google uses them. And it's kinda deprecated, afaik.
2. Unlike CBDT+CBLC/sbix, where we have just a PNG, here we have a custom encoding format. And the question is: should we decoded it or not. To my understanding, all EBDT bitmaps are 1bit, therefore they can be decoded into Planar8 quite easily. This would require some changes to the API so we could decode it on-demand and probably into user-provided buffer. How do you plan to use this data? Do you decode it yourself or do you need the actual raw data?

Also, what about [EBSC](https://learn.microsoft.com/en-us/typography/opentype/spec/ebsc)?

src/tables/cbdt.rs
```diff
@@ -52,28 +47,72 @@ impl<'a> Table<'a> {
...
+                    y: i16::from(metrics.y) - i16::from(metrics.height),
+                    width: u16::from(metrics.width),
+                    height: u16::from(metrics.height),
+                    pixels_per_em: location.ppem,
+                    format: match bit_depth {
+                        1 => RasterImageFormat::BitmapMonoPacked,
+                        2 => RasterImageFormat::BitmapGray2Packed,
+                        4 => RasterImageFormat::BitmapGray4Packed,
+                        8 => RasterImageFormat::BitmapGray8,
+                        32 => RasterImageFormat::BitmapBgra32,
```
RazrFalcon, line 101: Is it defined anywhere in the spec? I cannot find it. In the MS/OpenType spec, 32 isn't allowed. Is it OTB-specific?

src/tables/cblc.rs
```diff
@@ -45,24 +59,28 @@ fn select_bitmap_size_table(
...
         // Select a best matching subtable based on `pixels_per_em`.
-        if (pixels_per_em <= ppem && ppem < max_ppem)
-            || (pixels_per_em > max_ppem && ppem > max_ppem)
+        if (pixels_per_em <= ppem_x && ppem_x < max_ppem)
+            || (pixels_per_em > max_ppem && ppem_x > max_ppem)
         {
             idx = Some(usize::num_from(i));
-            max_ppem = ppem;
+            max_ppem = ppem_x;
+            max_bit_depth = bit_depth;
```
RazrFalcon, line 83: Is the last subtable expected to have the biggest bit depth?

src/tables/ebdt.rs
```diff
@@ -0,0 +1,30 @@
+//! An [Embedded Bitmap Data Table](
```
RazrFalcon, line 1: Since this is just a wrapper, we can use a typedef here. Just put it in CBDT and CBLC files respectively. No need for a new type.

tests/bitmap.rs
```diff
@@ -0,0 +1,66 @@
+use ttf_parser::{RasterGlyphImage, RasterImageFormat};
+
+static FONT_DATA: &[u8] = include_bytes!("fonts/bitmap.otb");
```
RazrFalcon, line 3: Add a comment that it's you own font and it has free license. CC0 I guess.

RazrFalcon (changes requested):

src/tables/cblc.rs
dzamkov: Nope, renamed to `bit_depth_for_max_ppem` to clarify.

tests/bitmap.rs
dzamkov: Done

src/tables/cbdt.rs
dzamkov: This is defined in the [CBDT ](https://learn.microsoft.com/en-us/typography/opentype/spec/cbdt) specs, under "Uncompressed Color Bitmaps"

RazrFalcon: Oh, so it's not from EBDT. Got it.

Also I would call it `BitmapPremulBgra32`

src/tables/ebdt.rs
dzamkov: These are public types that are exposed by the API. If we use a typedef, the user can rely on the fact that they are the same types and it would be a breaking change to separate the implementations. For instance, if we found some place where they need to be handled different. Are you sure you want this?

Also I'm a bit confused about what exactly you're asking for. Do you want the EBDT table type to be in the `cbdt` module? And if I did put it there, what would it be called, since the name `Table` is already taken?

RazrFalcon: Breaking changes are expected. This is not 1.0 library. No problem.

Oh yeah, typedef would not work. We need:

```rust
pub mod ebdt {
    pub use crate::cbdt::Table;
}
```

I thinks this should work.

Or maybe we can skip explicit EBDT+EBLC tables altogether? Can font have EBDT+EBLC and CBDT+CBLC at the same time? Probably not. In which case we can parse EBDT into CBDT and so on. And add a comment somewhere explaining the logic.

dzamkov:
To provide some context and answer your initial questions. I'm working on a game with a custom engine. I already implemented TrueType rasterization using your library, but I would also like to support bitmap fonts since they're relatively easy to make and appear "crisper" at lower resolutions. I've been browsing around for bitmap fonts, and it seems that OTB is a fairly common format to find them in, and is commonly supported for tools that export bitmap fonts. All the ones I've found use EBDT+EBLC. I haven't seen any CBDT, CBLC or EBSC tables.

For your second point, I don't think `ttf_parser` should attempt to decode image data, or at least, it should always provide access the raw data. There are so many ways to use it that it would be impossible to satisfy everyone otherwise. For example, I currently decode the data into a 16bpp buffer where the color is always 255, and the opacity comes from glyph data. I upload this to a texture on the GPU. At some point in the future I might switch to a 32bpp buffer where again, the color is always (255, 255, 255) and the opacity comes from the glyph data. Right now, I create an image per each glyph, but in the future I might create [one large image](https://en.wikipedia.org/wiki/Texture_atlas) that I decode all the glyphs into. Alternatively, for some formats (e.g. 8-bit grayscale and 32-bit bgra), it is possible to upload the raw data directly, skipping decoding entirely. I would not be happy if I had to allocate a temporary buffer that I didn't need.

Anyway, I wanted to thank you for this awesome library! It's been working great for me so far.

src/tables/ebdt.rs
dzamkov: That code wouldn't give us the right documentation, since it would still point to the CBDT specs.

My preference would be to keep things as is (which gives the correct documentation and doesn't expose implementation details). If not that, then quietly parse EBDT into CBDT as you proposed, but I have no idea whether a font can have both and if this is okay. Anyway, up to you.

src/tables/cbdt.rs
dzamkov: Updated

RazrFalcon:
I see. Thanks for the explanation.

To my understanding, the current implementation of EBDT+EBLC is identical to bdat+bloc. EBDT+EBLC added only formats 8 and 9, which we do not support anyway and which should be documented/mentioned somewhere. Including in the readme (it should be `~ (no 8, 9)`). And then we can mark bdat+bloc as fully supported.

As for the code, I would suggest adding:

```
pub bdat: Option<cbdt::Table<'a>>,
pub ebdt: Option<cbdt::Table<'a>>,
```

We already do the same trick with `hmtx` and `vmtx`. No need for separate tables/wrappers.

And in the `cbdt::Table` docs mention that this table is used to parse `bdat`, `EBDT` and `CBDT`.

PS: if you would have performance issues parsing those tables, we could pre-parse them more. Currently we're re-parsing a lot of data for each glyph.

src/lib.rs
```diff
@@ -431,6 +431,51 @@ impl OutlineBuilder for DummyOutline {
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
 pub enum RasterImageFormat {
     PNG,
+
+    /// A monochrome bitmap. The most significant bit of the first byte corresponds to the top-left
```
RazrFalcon, line 435: In Rust docs, the first line should be short, since this is a title. Then a blank one and then the rest of the docs.

Like so:

```rust
    /// A monochrome bitmap. 
    ///
    /// The most significant bit of the first byte corresponds to the top-left
    /// pixel, proceeding through succeeding bits moving left to right. The data for each row is
    /// padded to a byte boundary, so the next row begins with the most significant bit of a new
    /// byte. 1 corresponds to black, and 0 to white.
```

RazrFalcon:
Also update the c-api crate.

dzamkov:
Addressed the follow-up comments. From the [bdat specs](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6bdat.html), it doesn't allow formats 8 and 9, but it does allow format 4, so I marked that as unimplemented in the readme. I also marked 8 and 9 as unimplemented for CBDT because any format that is valid for EBDT is technically allowed to appear in CBDT.

I updated the C API but admittedly I didn't really know what I was doing and was just following patterns. You might need to adjust things.

README.md
```diff
@@ -60,17 +60,17 @@ There are roughly three types of TrueType tables:
 | Rendering         | -<sup>1</sup>          | ✓                   | ~ (very primitive)             |
 | `ankr` table      | ✓                      |                     |                                |
 | `avar` table      | ✓                      | ✓                   |                                |
-| `bdat` table      |                        | ✓                   |                                |
-| `bloc` table      |                        | ✓                   |                                |
-| `CBDT` table      | ✓                      | ✓                   |                                |
+| `bdat` table      | ~ (no 4)               | ✓                   |                                |
+| `bloc` table      | ✓                      | ✓                   |                                |
+| `CBDT` table      | ~ (no 8, 9)            | ✓                   |                                |
```
RazrFalcon, line 65: I guess no 4 here as well?

src/tables/cbdt.rs
```diff
@@ -1,12 +1,15 @@
 //! A [Color Bitmap Data Table](
 //! https://docs.microsoft.com/en-us/typography/opentype/spec/cbdt) implementation.
 
-use super::cblc::{self, BitmapFormat};
+use super::cblc;
+use crate::cblc::{Metrics, MetricsFormat, BitmapDataFormat};
```
RazrFalcon, line 5: Why not `use crate::cblc::{self, ..}` ?

RazrFalcon:
Thanks. Looks good to me.

README.md
dzamkov: I don't think so. The Microsoft specs say that it's only supported by Apple, but the Apple specs use bdat+bloc instead of EBDT+EBLC or CBDT+CBLC.

RazrFalcon: Ugh... ok. AAT and OT are indeed sometimes disagree on stuff.

src/tables/cbdt.rs
dzamkov: Because apparently that's what my IDE does when it auto-imports things and I don't double-check it.

RazrFalcon (approved):

RazrFalcon:
Thanks for you work! I will publish a new version soon.

RazrFalcon:
Published a new version.

Issue #69, opened 2021-10-22
Are there plans to support COLR/CPAL tables?


yisibl:
- COLRv0: https://docs.microsoft.com/en-us/typography/opentype/spec/colr
- COLRv1: https://docs.microsoft.com/en-us/typography/opentype/otspec190alpha/colr

COLR format fonts are part of the OpenType specification. It is the perfect companion for emoji and font icons in usage scenarios. This has become so popular that all major browsers support COLRv0.

![image](https://user-images.githubusercontent.com/2784308/138440311-82edf9b6-8406-438b-a272-e49c74558f8b.png)

And Chrome will soon [ship COLRv1](https://groups.google.com/a/chromium.org/g/blink-dev/c/kDfj3rcA6sc), which will bring many new features such as gradients, transforms and compositing.

2023-09-21

![image](https://github.com/RazrFalcon/ttf-parser/assets/2784308/ee4fd556-aff3-4a2e-9ea7-122dce76e0eb)

RazrFalcon:
Current progress: #133

PR #130, opened 2023-10-03
Add support for COLR/CPAL v0

laurmaedje:
This PR adds support for version 0 of the COLR and CPAL tables (#69). Version 0 is very simple in comparison to version 1: A color glyph is defined as a list of overlayed outline glyphs, each with a solid color.

The implementation is also pretty simple, the biggest question is the API design: Since a version 0 color glyph is conceptually really just a list of `(GlyphId, Color)`, an iterator based design would be possible. 

However, such an API design wouldn't accomodate version 1 down the road at all. For this reason, I decided to instead add a `ColorGlyphPainter` trait akin to the `OutlineBuilder` trait (I'm open for alternative naming suggestions). Currently, it only has a single method for painting an outline glyph in a color, but down the road it could be extended to accomodate other COLRv1 features like transformations and gradients. This will likely still be a breaking change (unless the new methods would have empty default impls), but at least the general API shape could remain the same.

I did not add CPAL to `FaceTables` for now, but it could potentially be useful to expose because a color font can have different palettes (e.g. light and dark) and it's up to a text processing application to pick one. CPAL v1 also adds support for named palettes, this especially might make it useful to expose in `FaceTables`.

### DIFF
--- a/src/tables/colr.rs
+++ b/src/tables/colr.rs
@@ -90,6 +90,13 @@ impl<'a> Table<'a> {
         })
     }
 
+    /// Whether the table contains a definition for the given glyph.
+    pub fn contains(&self, glyph_id: GlyphId) -> bool {
+        self.base_glyphs
+            .binary_search_by(|base| base.glyph_id.cmp(&glyph_id))
+            .is_some()
+    }
+
     /// Paints the color glyph.
     pub fn paint(
         &self,
--- a/src/tables/colr.rs
+++ b/src/tables/colr.rs
@@ -1,0 +1,115 @@
+//! A [Color Table](
+//! https://docs.microsoft.com/en-us/typography/opentype/spec/colr) implementation.
+
+use crate::cpal::{self, Color};
+use crate::parser::{FromData, LazyArray16, Offset, Offset32, Stream};
+use crate::GlyphId;
+
+/// A [base glyph](https://learn.microsoft.com/en-us/typography/opentype/spec/colr#baseglyph-and-layer-records).
+#[derive(Clone, Copy, Debug)]
+struct BaseGlyphRecord {
+    glyph_id: GlyphId,
+    first_layer_index: u16,
+    num_layers: u16,
+}
+
+impl FromData for BaseGlyphRecord {
+    const SIZE: usize = 6;
+
+    fn parse(data: &[u8]) -> Option<Self> {
+        let mut s = Stream::new(data);
+        Some(Self {
+            glyph_id: s.read::<GlyphId>()?,
+            first_layer_index: s.read::<u16>()?,
+            num_layers: s.read::<u16>()?,
+        })
+    }
+}
+
+/// A [layer](https://learn.microsoft.com/en-us/typography/opentype/spec/colr#baseglyph-and-layer-records).
+#[derive(Clone, Copy, Debug)]
+struct LayerRecord {
+    glyph_id: GlyphId,
+    palette_index: u16,
+}
+
+impl FromData for LayerRecord {
+    const SIZE: usize = 4;
+
+    fn parse(data: &[u8]) -> Option<Self> {
+        let mut s = Stream::new(data);
+        Some(Self {
+            glyph_id: s.read::<GlyphId>()?,
+            palette_index: s.read::<u16>()?,
+        })
+    }
+}
+
+/// A trait for color glyph painting.
+pub trait ColorGlyphPainter {
+    /// Paints an outline glyph in the given color.
+    fn glyph(&mut self, id: GlyphId, color: Color);
+}
+
+/// A [Color Table](
+/// https://docs.microsoft.com/en-us/typography/opentype/spec/colr).
+///
+/// Currently, only version 0 is supported.
+#[derive(Clone, Copy, Debug)]
+pub struct Table<'a> {
+    palettes: cpal::Table<'a>,
+    base_glyphs: LazyArray16<'a, BaseGlyphRecord>,
+    layers: LazyArray16<'a, LayerRecord>,
+}
+
+impl<'a> Table<'a> {
+    /// Parses a table from raw data.
+    pub fn parse(palettes: cpal::Table<'a>, data: &'a [u8]) -> Option<Self> {
+        let mut s = Stream::new(data);
+
+        let version = s.read::<u16>()?;
+        if version > 1 {
+            return None;
+        }
+
+        let num_base_glyphs = s.read::<u16>()?;
+        let base_glyphs_offset = s.read::<Offset32>()?;
+        let layers_offset = s.read::<Offset32>()?;
+        let num_layers = s.read::<u16>()?;
+
+        let base_glyphs = Stream::new_at(data, base_glyphs_offset.to_usize())?
+            .read_array16::<BaseGlyphRecord>(num_base_glyphs)?;
+
+        let layers = Stream::new_at(data, layers_offset.to_usize())?
+            .read_array16::<LayerRecord>(num_layers)?;
+
+        Some(Self {
+            palettes,
+            base_glyphs,
+            layers,
+        })
+    }
+
+    /// Paints the color glyph.
+    pub fn paint(
+        &self,
+        glyph_id: GlyphId,
+        palette: u16,
+        painter: &mut dyn ColorGlyphPainter,
+    ) -> Option<()> {
+        let (_, base) = self
+            .base_glyphs
+            .binary_search_by(|base| base.glyph_id.cmp(&glyph_id))?;
+
+        let start = base.first_layer_index;
+        let end = start.checked_add(base.num_layers)?;
+        let layers = self.layers.slice(start..end)?;
+
+        for layer in layers {
+            let color = self.palettes.get(palette, layer.palette_index)?;
+            painter.glyph(layer.glyph_id, color);
+        }
+
+        Some(())
+    }
+}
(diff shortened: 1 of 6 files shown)


RazrFalcon:
Nice! Thanks!

I haven't looked much into COLR, but to my understanding it's a very simple SVG alternative. 
As for the API, I would prefer an iterator. But I'm not sure it would be able to handle v1 that well.
I would suggest Implementing v1 as well, so we would not need to break the API twice.

I would try to take a look into COLR v1 on the weekeds to have any meaningful comments.

As for `FaceTables`, all tables must be present in it. So feel free to add it. It does skip "unnecessary" tables like `loca`, but that's about it.

Also, even some rudimentary tests would be welcome. 
And also, were can I find fonts with COLR tables? I know only Segoe UI Emoji by Microsoft and [mozilla/twemoji-colr](https://github.com/mozilla/twemoji-colr).

laurmaedje:
> I haven't looked much into COLR, but to my understanding it's a very simple SVG alternative.

COLRv0 is super simple, while COLRv1 is described as being similar to SVG Native. It has gradients, transforms, blend modes and a few other things.

> As for FaceTables, all tables must be present in it. So feel free to add it. It does skip "unnecessary" tables like loca, but that's about it.

Done!

> Also, even some rudimentary tests would be welcome.

I could hand-code COLR and CPAL v0 tables for a simple test. Or should I use the `demo.ttx` infrastructure? I'm not sure whether it supports COLR.

> As for the API, I would prefer an iterator. But I'm not sure it would be able to handle v1 that well.

V1 is described by a directed acyclic graph of paint tables. I think it would be quite difficult to expose as an iterator, but maybe it's possible. I haven't looked at its details.

> And also, were can I find fonts with COLR tables?

I only know Segoe UI Emoji, didn't even know about twemoji-colr. I justed updated my Windows PC to the newest Windows version (which ships with a 'new' Segoe UI Emoji design). Even then, the font still uses COLR v0. I'm not aware of any COLR v1 font. In principle, it would of course be nice to support it, but I wouldn't know how to test it and I'm also probably too stretched for time to implement it fully.

RazrFalcon:
>I could hand-code COLR and CPAL v0 tables for a simple test. Or should I use the demo.ttx infrastructure? I'm not sure whether it supports COLR.

Just see how current tests are done. It's just a commented binary blob. Pain to write, but hey, that's TTF.
`demo.ttx` is for different stuff.

>In principle, it would of course be nice to support it, but I wouldn't know how to test it and I'm also probably too stretched for time to implement it fully.

Well, that's kinda the point. If no one uses it - why we have to implement it? And without having a real-world font it's basically impossible to test. The spec alone is not enough. 

I guess we can merge v0 as long as it works fine from your side and leave v1 for later.

laurmaedje:
> Just see how current tests are done. It's just a commented binary blob. Pain to write, but hey, that's TTF.

Okay, I'll do that.

> I guess we can merge v0 as long as it works fine from your side and leave v1 for later.

It works for my use case (rendering color emojis in Typst and exporting them to PDF), I tested it integrated with Typst. I don't have a strong opinion on whether to keep the trait-based API (in anticipation of it being useful later) or switch to an iterator (better API for v0). I would leave that up to you.

anthrotype:
> If no one uses it

we have published a bunch of COLRv1 fonts on https://fonts.google.com/?coloronly=true
Most notably, [Noto Color Emoji](https://fonts.google.com/noto/specimen/Noto+Color+Emoji?coloronly=true). 

You can build your own COLRv0/v1 test fonts from a bunch of SVG files using the [nanoemoji](https://github.com/googlefonts/nanoemoji) color font compiler

RazrFalcon:
@anthrotype Nice. Thanks!

@laurmaedje I think I will merge you code now and will take a look at v1 on weekends and then decide if we should release it as is or I will add v1 myself.

laurmaedje:
> Most notably, [Noto Color Emoji](https://fonts.google.com/noto/specimen/Noto+Color+Emoji?coloronly=true).

Interesting, looks like it includes _both_ COLR and SVG tables.

> I think I will merge you code now and will take a look at v1 on weekends and then decide if we should release it as is or I will add v1 myself.

laurmaedje:
> When serving them as webfonts we pick the right table and subset the other away

That's pretty cool.

RazrFalcon:
>Should I still add the tests in a separate PR?

If you have time - sure.

laurmaedje:
> If you have time - sure.

I don't have much, but if I say I'm gonna create a test, I'll stick to it. :) https://github.com/RazrFalcon/ttf-parser/pull/131

yisibl:
> And also, were can I find fonts with COLR tables?

`iconfont.cn` supported the COLR v0 font format a few years ago, which has been widely used for icons.

Here's an [introductory article](https://www.iconfont.cn/help/article_detail?spm=a313x.help_detail.i1.daea9c966.77883a81niNDz5&article_id=7) (sorry it's in Chinese).

yisibl:
Thank you @laurmaedje for the amazing work.

src/tables/colr.rs
```diff
@@ -0,0 +1,122 @@
...
+        self.base_glyphs
+            .binary_search_by(|base| base.glyph_id.cmp(&glyph_id))
+            .is_some()
+    }
+
+    /// Paints the color glyph.
+    pub fn paint(
+        &self,
+        glyph_id: GlyphId,
+        palette: u16,
```
RazrFalcon, line 104: @laurmaedje From where the `palette` values comes from?

laurmaedje: The CPAL table has one or multiple palettes and this value selects which one. 0 will always work. I should probably have added a function to CPAL that gives you the number of palettes. It's not too useful without the palette names though and this is a CPALv1 feature I intentionally skipped.

This value would only be something other than zero if an application can make sense of multiple color palettes in one font, e.g. by exposing it to the user or picking a darker palette in dark mode.

RazrFalcon: We're talking about `CPAL.numPalettes` right? I think I get it now. I will update the docs and the API.

PS: Currently trying to add COLR support to `font2svg` example. Seems to be working fine. But we do have to provide some API to check if the outline is just "a layer" and not a stand-alone glyph.

laurmaedje: > We're talking about `CPAL.numPalettes` right? I think I get it now. I will update the docs and the API.

Yes, numPalettes is what I meant.

> But we do have to provide some API to check if the outline is just "a layer" and not a stand-alone glyph.

I think technically a glyph could be both. For the purposes of the example, you could maybe use the cmap.codepoints API in combination with glyph_index to find all glyphs that are referenced by the cmap table.

RazrFalcon: As always, everything is far more complicated than it should be.

1. I still have no idea how to check if a glyph is a base glyph,  a color layer glyph or just a normal outline. Checking all layers in COLR would be very expensive.
2. We have to somehow handle 0xFFFF palette. "A paletteIndex value of 0xFFFF is a special case, indicating that the text foreground color (as determined by the application) is to be used." Meaning that even for v0 we have to have `Painter::foreground`

laurmaedje: 1. Why do you want to check that in the first place? I think it's not designed to be checked for.

2. Oh, that's my bad. I think the painter callback could just give an `Option<Color>` where `None` means foreground?

RazrFalcon: 1. Mainly for `font2svg`. But I guess in the normal use case it wouldn't be needed.

laurmaedje: For font2svg it's maybe not that big of a deal if it's expensive.

RazrFalcon:
Well, v1 is pretty crazy. Parsing is ok-ish, but we do need a different API. Still trait-based, but different. Glyph outlining and color filling had to be separated.
Also, v1 isn't very compatible with SVG, therefore testing is troublesome. I have no plans on writing the whole renderer for tests.
And v1 is also variable, which adds a whole another level of complexity.
I did wrote most of v1 without variation support, but I do not think I would be able to finish it this week.
@laurmaedje I think we can release v0 with an API designed for v1 and then add v1 later, if you're looking for a quick release.

RazrFalcon:
@anthrotype Hi! Are there any variable COLR fonts out there?

laurmaedje:
@RazrFalcon Releasing a compatible v0 first seems sensible to me. That said, I'm not in too much of a hurry, so if we're talking just a few weeks I can also wait. That's up to you to decide. In any case, thanks for working on this!

yisibl:
> Hi! Are there any variable COLR fonts out there?

https://nabla.typearture.com/

RazrFalcon:
@yisibl Those are variable fonts, but not COLRv1 variable. I need a font that uses COLRv1 variations.

yisibl:
@RazrFalcon Nabla is a variable font for COLRv1.

RazrFalcon:
@yisibl Nable has variable glyphs, but not the variable COLRv1. Those are different.

yisibl:
@RazrFalcon Oh sorry, please see if this is what you need: https://roettsch.es/var_colrv1.html

RazrFalcon:
Seems like it. But where can I get the font itself?

anthrotype:
from here: https://github.com/googlefonts/color-fonts

anthrotype:
this is a variable COLRv1 font as well: https://fonts.google.com/specimen/Foldit

RazrFalcon:
@anthrotype Thanks! Foldit is what I was looking for.

anthrotype:
though Foldit might now exercise all the COLRv1 Paint formats, the other test font that @yisibl suggested (which is also used for skia test suite) has better coverage

RazrFalcon:
Which font is that?

yisibl:
@RazrFalcon Is here: https://github.com/googlefonts/color-fonts/blob/main/fonts/test_glyphs-glyf_colr_1_variable.ttf

The author is @drott

RazrFalcon:
Thanks.

PR #132, opened 2023-10-06
return the svg document include the svg data and the covered range 

wjian23:
A single svg document can be used to represent multiple glyphs, return the covered range with the data.

### DIFF
--- a/src/tables/svg.rs
+++ b/src/tables/svg.rs
@@ -3,6 +3,17 @@
 use crate::parser::{FromData, LazyArray16, NumFrom, Offset, Offset32, Stream};
 use crate::GlyphId;
 
+#[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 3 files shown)


src/tables/svg.rs
```diff
@@ -3,6 +3,17 @@
 use crate::parser::{FromData, LazyArray16, NumFrom, Offset, Offset32, Stream};
 use crate::GlyphId;
 
+#[derive(Debug, Clone, Copy)]
+/// An SvgDocument can include multi glyphs
+pub struct SvgDocment<'a> {
```
RazrFalcon, line 8: SvgDoc**u**ment

wjian23: 👍，typo fixed

RazrFalcon:
Thanks.

PR #133, opened 2023-10-08
Support COLR v1 table.

RazrFalcon:
TODO:

- [ ] ClipList table
- [ ] Variations
- [ ] PaintColrGlyph
- [ ] Recursion detection
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -483,6 +483,33 @@ impl core::fmt::Debug for Transform {
     }
 }
 
+/// A RGBA color in the sRGB color space.
+#[allow(missing_docs)]
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub struct RgbaColor {
+    pub red: u8,
+    pub green: u8,
+    pub blue: u8,
+    pub alpha: u8,
+}
+
+impl RgbaColor {
+    /// Creates a new `RgbaColor`.
+    #[inline]
+    pub fn new(red: u8, green: u8, blue: u8, alpha: u8) -> Self {
+        Self {
+            blue,
+            green,
+            red,
+            alpha,
+        }
+    }
+
+    pub(crate) fn apply_alpha(&mut self, alpha: f32) {
+        self.alpha = (((f32::from(self.alpha) / 255.0) * alpha) * 255.0) as u8;
+    }
+}
+
 /// A trait for glyph outline construction.
 pub trait OutlineBuilder {
     /// Appends a MoveTo segment.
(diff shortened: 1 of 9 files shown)


RazrFalcon:
@laurmaedje beta quality, but seems to be working ok-ish. Here is `font2svg` output:

sample

Some stuff is hard to represent in SVG, like weird transforms order, sweep gradients, gradients rotation (?), some blending modes.

For you this just a progress report and you can try out the new API. 
PS: I've added `colr::Table::is_simple` to reject v1 tables in case you do not plan to support them yet.

src/lib.rs
```diff
@@ -405,6 +405,84 @@ impl BBox {
...
+    pub f: f32,
+}
+
+impl Transform {
+    #[inline]
+    pub fn new(a: f32, b: f32, c: f32, d: f32, e: f32, f: f32) -> Self {
+        Transform { a, b, c, d, e, f }
+    }
+
+    #[cfg(feature = "variable-fonts")]
```
laurmaedje, line 424: this cfg is a bit strange

src/tables/colr.rs
```diff
@@ -49,13 +51,325 @@ impl FromData for LayerRecord {
...
+    Pad,
+    Repeat,
+    Reflect,
+}
+
+impl FromData for GradientExtend {
+    const SIZE: usize = 1;
+
+    fn parse(data: &[u8]) -> Option<Self> {
+        match data[0] {
```
laurmaedje, line 87: won't panic in practice due to where it's called I assume, but still an unfamiliar sight in this crate since it can technically panic. any reason to prefer this over `data.first()?`?

laurmaedje (commented):
Nice work! There are two minor things I noticed while reading.

src/lib.rs
RazrFalcon: Yes, we should either make it private, as it was before, or remove cfg altogether.

src/tables/colr.rs
RazrFalcon: It's pretty common. `data` is guarantee to be `Self::SIZE`.

laurmaedje: Okay, good to know.

yisibl:
> Some stuff is hard to represent in SVG, like weird transforms order, sweep gradients, gradients rotation (?), some blending modes.

About `sweep gradients`: https://github.com/Lorp/samsa-core/issues/3

yisibl:
Maybe this helps, FYI. https://github.com/googlefonts/oxidize/pull/48

RazrFalcon:
It's a bit different, but thanks.

RazrFalcon:
@laurmaedje I've published a new ttf-parser with COLR v0. No timeline for v1.

LaurenzV:
If you think you don't have time to finish it, I can look into finishing this in the next few weeks.

RazrFalcon:
Yeah, no plans for this month. And it's a tricky beast.

Maybe would have time to at least update the [ttf-explorer](https://github.com/RazrFalcon/ttf-explorer).

[Test font](https://github.com/googlefonts/color-fonts/blob/main/fonts/test_glyphs-glyf_colr_1_variable.ttf).

LaurenzV:
Do you know if there is also some way I can view the "reference" of how each glyph is supposed to look to compare? Something like `font2svg`, but where we know that it (should) already work correctly.

LaurenzV:
Okay, looks like `hb-view` supports it. I think I should be able to work with that.

RazrFalcon:
I was using Chrome.

LaurenzV:
Could you tell me how exactly you used Chrome to test it? So how did you make it show the glyphs that you wanted to test?

RazrFalcon:
Via this page: https://fonts.google.com/noto/specimen/Noto+Color+Emoji/glyphs

RazrFalcon:
Superseded by #145

PR #135, opened 2023-12-13
Proposing nit edits to project's description

Leedehai:
Hi I'd like to propose nit edits to the README.md and Cargo.toml file to clarify up front that the library is also capable of parsing OpenType and AAT, in addition to TrueType.

I landed on ttf-parser's directory when browsing [Typst](https://github.com/typst/typst)'s codebase following Rust Analyzer's cross reference links, and was bewildered why a TrueType font parser (that was what I understood "ttf" in the name as) can handle OpenType files, so I spent quite a few moments searching for an OpenType parser in Typst's dependency list, before finally noticing the second paragraph in ttf-parser's README.md that mentions OpenType. Therefore, I'd like to humbly propose these edits to make it obvious :)

Finally, this is an impressive piece of work! Thank you for making it:)


RazrFalcon:
Thanks. Makes sense. 
I know that the library name is a bit confusing to people, but OpenType and AAT _are_ TrueType. And the file extension doesn't matter much either.

PR #137, opened 2023-12-24
Flip misnamed `is_bitmap_embedding_allowed` function, increase lenience of embed permissions for older OS/2 versions

Fuzzyzilla:
After referencing [this Microsoft `fsType` specification](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#fst), I noticed an inconsistency in the functions `{Face, os2::Table}::is_bitmap_embedding_allowed` which seemed to have an inverted interpretation of the flag's meaning. The checked flag, if set, really means "is embed *required* to be bitmap". Since that name isn't very clear either, I inverted the name to `is_outline_embedding_allowed` while keeping the logical return value unflipped.

Further, the specification allows much leniency for earlier versions of the OS/2 table, this PR adjusts the logic of `os2::Table::{permissions, is_subsetting_allowed, is_outline_embedding_allowed}` to take advantage of this to be as lenient as allowable for embeds.

After these changes, I am not sure the default values for when the table isn't found are correct. I assumed that, in a version prior to the flags' introduction, both subsetting and outline embedding are allowed due to the negative naming of the flags (eg. "no subsetting" implies subsetting is the default) but I was unable to find a resource that states this outright.

### DIFF
--- a/src/tables/os2.rs
+++ b/src/tables/os2.rs
@@ -433,16 +433,19 @@ impl<'a> Table<'a> {
         }
     }
 
-    /// Checks if the face subsetting is allowed.
+    /// Checks if the face allows embedding a subset, further restricted by [`Self::permissions`].
     #[inline]
     pub fn is_subsetting_allowed(&self) -> bool {
         let n = Stream::read_at::<u16>(self.data, TYPE_OFFSET).unwrap_or(0);
         n & 0x0100 == 0
     }
 
-    /// Checks if the face bitmaps embedding is allowed.
+    /// Checks if the face allows outline data to be embedded.
+    /// If false, only bitmaps may be embedded in accordance with [`Self::permissions`].
+    ///
+    /// If the font contains no bitmaps and this flag is not set, it implies no embedding is allowed.
     #[inline]
-    pub fn is_bitmap_embedding_allowed(&self) -> bool {
+    pub fn is_outline_embedding_allowed(&self) -> bool {
         let n = Stream::read_at::<u16>(self.data, TYPE_OFFSET).unwrap_or(0);
         n & 0x0200 == 0
     }
--- a/src/tables/os2.rs
+++ b/src/tables/os2.rs
@@ -424,20 +424,42 @@ impl<'a> Table<'a> {
     #[inline]
     pub fn permissions(&self) -> Option<Permissions> {
         let n = Stream::read_at::<u16>(self.data, TYPE_OFFSET).unwrap_or(0);
-        match n & 0xF {
-            0 => Some(Permissions::Installable),
-            2 => Some(Permissions::Restricted),
-            4 => Some(Permissions::PreviewAndPrint),
-            8 => Some(Permissions::Editable),
-            _ => None,
+        if self.version <= 2 {
+            // Version 2 and prior, applications are allowed to take
+            // the most permissive of provided flags
+            let permission = if n & 0xF == 0 {
+                Permissions::Installable
+            } else if n & 8 != 0 {
+                Permissions::Editable
+            } else if n & 4 != 0 {
+                Permissions::PreviewAndPrint
+            } else {
+                Permissions::Restricted
+            };
+
+            Some(permission)
+        } else {
+            // Version 3 onwards, flags must be mutually exclusive.
+            match n & 0xF {
+                0 => Some(Permissions::Installable),
+                2 => Some(Permissions::Restricted),
+                4 => Some(Permissions::PreviewAndPrint),
+                8 => Some(Permissions::Editable),
+                _ => None,
+            }
         }
     }
 
     /// Checks if the face allows embedding a subset, further restricted by [`Self::permissions`].
     #[inline]
     pub fn is_subsetting_allowed(&self) -> bool {
-        let n = Stream::read_at::<u16>(self.data, TYPE_OFFSET).unwrap_or(0);
-        n & 0x0100 == 0
+        if self.version <= 1 {
+            // Flag introduced in version 2
+            true
+        } else {
+            let n = Stream::read_at::<u16>(self.data, TYPE_OFFSET).unwrap_or(0);
+            n & 0x0100 == 0
+        }
     }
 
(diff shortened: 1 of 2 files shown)


src/tables/os2.rs
```diff
@@ -424,27 +424,57 @@ impl<'a> Table<'a> {
...
+    /// Checks if the face allows embedding a subset, further restricted by [`Self::permissions`].
     #[inline]
     pub fn is_subsetting_allowed(&self) -> bool {
-        let n = Stream::read_at::<u16>(self.data, TYPE_OFFSET).unwrap_or(0);
-        n & 0x0100 == 0
+        if self.version <= 1 {
+            // Flag introduced in version 2
+            true
+        } else {
+            let n = Stream::read_at::<u16>(self.data, TYPE_OFFSET).unwrap_or(0);
```
RazrFalcon, line 460: Wouldn't we always return true for old versions because of `unwrap_or(0)`? I agree that your code is more readable, just clarifying the logic.

RazrFalcon:
I personally have no idea how it should work. And it's hard to test. Will see how CoreText handles it.

src/tables/os2.rs
Fuzzyzilla: Old versions (specifically versions 0 and 1) neglected to specify that unused bits are zeroed, so they could have any value: "Applications must ignore bits 4 to 15 when reading a version 0 or version 1 table."
While unlikely any fonts in the wild actually did write non-zeroes here, might as well be strict with the implementation x3

RazrFalcon: Oh right, my bad. I thought we're slicing the exact length, therefore read out of bound would always return 0.

Fuzzyzilla:
I have checked FreeType, fontTools, and FontForge for their behaviour - forgive me if I am missing obvious ones, I am not a usual visitor to the realm of font parsing!

* FreeType - the `fsType` flags are not parsed directly, instead yielded to the user as-is along with a [bunch of bitflags](https://github.com/freetype/freetype/blob/ca76683b781db5d06ef1a0e2cb62a767e7dbe626/include/freetype/freetype.h#L4657) to check against. Documentation does not hint at the version differences, though no relevant behavior since it's up to user code to determine how to handle old versions of the flags.
* fontTools - when merging fonts, the no-subset/bitmap-only bits are accessed but merely by copying over to the merge result instead of choosing the strictest. In particular though, for versions 0 thru 2 without the flag, [it sets them to zero (most permissive)](https://github.com/fonttools/fonttools/blob/2dc887c894e6cd81d91f8dd13c919a3cdf616118/Lib/fontTools/merge/tables.py#L171)
* FontForge - writes and parses the no-subset/bitmap-only bits regardless of OS/2 version (checked by experimentation)

In general, it seems that the support of these flags is a bit hit-or-miss in terms of adherence to the Microsoft spec I posted above!

RazrFalcon:
I took a closer look and indeed `is_bitmap_embedding_allowed` is flipped. Probably just copy-pasted `is_subsetting_allowed`.

As for other apps, I couldn't find a similar API in CoreGraphics/CoreText. Font Book on macOS and Explorer on Windows do show font's permissions, but I don't have a font with the bitmap bit set. Do you have any?

> I am not a usual visitor to the realm of font parsing!

Well, then welcome to a barely documented mess. You might be overwhelmed by how much Microsoft spec documents, but it's just a tip of an iceberg.

Fuzzyzilla:
I have made up some barebones test fonts with the relevant flags set [here.](https://drive.google.com/drive/folders/1QVx34Oy30VqsFm-dusxDlgBff7lxvvBL?usp=sharing) (I figured it would be legally dubious to set these flags on an actual font i just grabbed somewhere :P )
They are both OS/2 Version 4 despite my best efforts to limit them to a lower version, but hopefully this helps!

If rather you meant had I ever encountered these fonts in the wild, I can't say I have - I am just in the process of writing a software which may need to perform font embedding and I would like to respect the licensing rules of any font the user may import. On my current system, all of my fonts are FOSS, so certainly none there have such restriction for me to test!

RazrFalcon:
Interesting. Seems like Apple recognizes this flag. Here is what Font Book.app shows:

`TestFontNoSubsetNoOutlines.ttf`: 

>No embedding restrictions. The font may not be subsetted prior to embedding. Only bitmaps contained in the font may be embedded. No outline data may be embedded.

`TestFontNoSubset.ttf`:

>No embedding restrictions. The font may not be subsetted prior to embedding. 

`TestFontNoOutlines.ttf`:

>No embedding restrictions. Only bitmaps contained in the font may be embedded. No outline data may be embedded. 

and just in case, here is _Apple Color Emoji_ for the reference: 

>Preview and print embedding. This font may be embedded in documents and temporarily loaded on the remote system. Documents containing this font must be opened “read-only;” no edits can be applied to the document.

Fuzzyzilla:
I managed to get FontForge to export with OS/2 version 1 after much fiddling (not sure what i did to make it happen, the version dropdown doesn't actually work!) [Please give it a test :3](https://drive.google.com/file/d/1xV9gv6oEMuvW8CB7nB-xr_3FnZS-_d7X/view?usp=drive_link)

Notably, I set the flags even when they're invalid to set on this version. FontForge reads them regardless, as does current `master`, curious to see the behavior of your font viewer apps in this situation.

RazrFalcon:
>No embedding restrictions. The font may not be subsetted prior to embedding. Only bitmaps contained in the font may be embedded. No outline data may be embedded.

So the same as `TestFontNoSubsetNoOutlines.ttf`.

Fuzzyzilla:
So I guess the question now is whether behavior should match the spec to the letter, or follow what the rest of the ecosystem does - it seems most others parse and use the bits regardless of OS/2 version, as current master does. My purpose for following the spec was to achieve maximum leniency - allow embedding wherever possible, and sealing up those little edge-cases to make it more likely that embedding was allowed while still following the letter of the law :3

RazrFalcon:
I think we can keep checking OS/2 version. It doesn't add much complexity and not many fonts still use old OS/2 versions.

PR #138, opened 2024-01-17
Make `LayoutTable::parse()` public

MoSal:
Will be needed for raw gpos/gsub parsing.


MoSal:
Not needed following the advice RazrFalcon/fontdb#62. Closing.

PR #140, opened 2024-02-07
minor refactor: use built-in option methods instead of `try_for_opt!` macro

SKalt:
I was reading through the code and had to check what `try_for_opt!` did. I factored the macro into built-in option methods to improve my comprehension of the code. I recognize this is a minor refactor that wasn't requested; please feel free to close this PR if it's too minor or it decreases readability.

In any case, thanks for writing `ttf-parser`! I'm enjoying using it.

### DIFF
--- a/src/tables/gvar.rs
+++ b/src/tables/gvar.rs
@@ -1626,8 +1626,13 @@ fn infer_delta(
         //
         // 'Target point delta is derived from the adjacent point deltas
         // using linear interpolation.'
-        let d = f32::from(try_opt_or!(target_point.checked_sub(prev_point), 0.0))
-            / f32::from(try_opt_or!(next_point.checked_sub(prev_point), 0.0));
+        let target_sub = target_point.checked_sub(prev_point);
+        let next_sub = next_point.checked_sub(prev_point);
+        let d = if let (Some(target_sub), Some(next_sub)) = (target_sub, next_sub) {
+            f32::from(target_sub) / f32::from(next_sub)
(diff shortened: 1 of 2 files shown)


src/tables/gvar.rs
```diff
@@ -1626,8 +1626,13 @@ fn infer_delta(
...
         // using linear interpolation.'
-        let d = f32::from(try_opt_or!(target_point.checked_sub(prev_point), 0.0))
-            / f32::from(try_opt_or!(next_point.checked_sub(prev_point), 0.0));
+
+        let target_sub = target_point.checked_sub(prev_point);
+        let next_sub = next_point.checked_sub(prev_point);
+        if target_sub.or(next_sub).is_none() {
+            return 0.0;
+        };
+        let d = f32::from(target_sub.unwrap()) / f32::from(next_sub.unwrap());
```
RazrFalcon, line 1635: I would prefer:

```rust
let d = if let (Some(target_sub), Some(next_sub)) = (target_sub, next_sub) {
    f32::from(target_sub) / f32::from(next_sub)
} else {
    return 0.0
};
```

RazrFalcon:
Looks like you have removed your `checked_sub` lines as well. They should stay.

SKalt:
Whoops. I fixed that and rebased my changes into a single commit, which I re-checked with `cargo check` and `cargo fmt` locally.

RazrFalcon:
All good. Thanks!

PR #145, opened 2024-04-24
Continue support COLRv1 table

LaurenzV:


LaurenzV:
@RazrFalcon As far as I can tell, the current way we deal with outlines is not going to be good enough to work with COLRv1 correctly... In particular, filling a shape must also create a clip path, so that if we have nested fills, the clip paths are combined. See also https://learn.microsoft.com/en-us/typography/opentype/spec/colr#filling-shapes. Currently, we only have the `outline` function which takes a glyph ID and the user must generate the path for it and store it until the next operation. 

My suggestions would be the following: Instead of having a single `outline` function, we implement methods for building paths (similar to `OutLineBuilder`) like `move_to`, `line_to` etc. and in the end, and then we also add a `clip`, `fill` and `clip_and_fill` method where the user should take the path that has been built and apply the corresponding operation. This would also allow us to define arbitrary clip paths, which I think will be necessary for `ClipBox`.

What do you think?

RazrFalcon:
Sure. Sounds good.

LaurenzV:
Small update: Making decent progress, but some parts of it are a bit tricky, especially dealing with transforms that only apply to a paint and not to an outline... But I'll think of something and make the API a bit nicer while I'm at it.

Only major thing missing is variable fonts, not sure how difficult that will be to implement.

LaurenzV:
https://github.com/RazrFalcon/ttf-parser/blob/ad4449de309b06d27a0d36ad48f75c5ad6711001/src/tables/colr.rs#L364-L366

What exactly is the "Explain why" comment referring to? And could we merge all of the `scale`, `rotate` etc. functions into `transform` or is there a good reason to keep them separate?

RazrFalcon:
>What exactly is the "Explain why" comment referring to? 

I already forgot 😄 

I would guess this is because rotate to affine transform requires trigonometry functions, which are not available in `no_std`. And we do not support `libm` hack in `ttf-parser` (yet).

I don't mind having just the `transform` callback, but we would have to port the `libm` hack from `tiny-skia`.

LaurenzV:
I'm getting there. Only documentation, cleanup and tests missing now.

RazrFalcon:
Glad to hear!

src/lib.rs
```diff
@@ -2270,30 +2316,6 @@ impl<'a> Face<'a> {
...
-                let end = offset
-                    .checked_add(length)
-                    .ok_or(FaceParsingError::MalformedFont)?;
-                let range = offset..end;
-                self.data.get(range)
-            }))
-        })
-    }
-}
-
```
LaurenzV, line 2296: Dead code.

src/lib.rs
```diff
@@ -658,15 +674,25 @@ impl FromData for TableRecord {
 }
 
 #[cfg(feature = "variable-fonts")]
-const MAX_VAR_COORDS: usize = 32;
+const MAX_VAR_COORDS: usize = 64;
```
LaurenzV, line 677: Had to increase it because the test font has more than 32 axes. :( I hope that's okay.

src/parser.rs
```diff
@@ -781,11 +795,6 @@ impl<'a> Stream<'a> {
 pub trait Offset {
     /// Converts the offset to `usize`.
     fn to_usize(&self) -> usize;
-
-    /// Checks that offset is null.
-    fn is_null(&self) -> bool {
-        self.to_usize() == 0
-    }
```
LaurenzV, line 788: Dead code.

src/tables/colr.rs
```diff
@@ -47,16 +185,523 @@ impl FromData for LayerRecord {
...
+    }
+}
+
+impl<'a> SweepGradient<'a> {
+    /// An iterator over the stops of the sweep gradient.
+    // TODO: Make API nicer (same for radial and linear gradient)
+    pub fn stops<'b>(
+        &'b self,
+        palette: u16,
+        #[cfg(feature = "variable-fonts")] coords: &'a [NormalizedCoordinate],
```
LaurenzV, line 512: It's not so nice that you have to manually pass the coords for this, instead of them being taken directly from the face. i tried it without it but it turned into a bit of a lifetime mess. So I will leave this as a potential improvement for the future.

src/var_store.rs
```diff
@@ -80,36 +80,55 @@ impl<'a> ItemVariationStore<'a> {
...
 
         let mut delta = 0.0;
         let mut i = 0;
-        while i < short_delta_count {
+        while i < word_delta_count {
             let idx = region_indices.get(i)?;
-            delta += f32::from(s.read::<i16>()?) * self.regions.evaluate_region(idx, coordinates);
+            let num = if has_long_words {
+                // TODO: use f64?
+                s.read::<i32>()? as f32
```
LaurenzV, line 107: Not sure if we should switch to f64 here, because in theory it could overflow...

LaurenzV:
Alright, I think it should be mostly done. Also left some comments (out of which only one of them is really a question).

Regarding testing, this is what I did:
- I converted the fonts with `font2svg` and comapred the output with `hb-draw`. There were quite a few differences, but all of them are due to limitations of SVG (e.g. all sweep gradient test cases were blank, some gradients looked different because they work differently, some blend modes don't work, etc.), but in all cases where there were differences I made sure that `ttf-parser` is still correct and it's just the SVG output that is wrong.
- For the static tests, I mostly relied on `fonttools ttx` output as well as the comparison to `hb-draw` to determine whether a test case is correct. For variation tests, I cross-compared all results with `fontations` to make sure that setting the variation axes leads to the same result. And this was always the case, apart from some very small rounding errors in some cases.

I hope it's good.

LaurenzV:
(Also, I'm not sure how CI works, before it ran automatically but now it says I need approval...)

LaurenzV:
Okay, CI seems to run. I had to bump CI to Rust 1.51 for const generics to work, I hope that's okay.

src/lib.rs
RazrFalcon: I hope so as well... Should fine I guess.

src/tables/colr.rs
RazrFalcon: Sure. No problem.

src/var_store.rs
RazrFalcon: It's fine for now. Let's leave it for the future.

Cargo.toml
```diff
@@ -36,6 +36,7 @@ glyph-names = []
 gvar-alloc = ["std"]
 
 [dev-dependencies]
-base64 = "0.13"
+base64 = "0.22.1"
 pico-args = "0.5"
+tiny-skia-path = "0.11.4"
```
RazrFalcon, line 41: If we need it just for path transform then it can be easily implemented without any dependencies. [See this code](https://github.com/RazrFalcon/tiny-skia/blob/9b673e6c14e5e0c3f98413f8fa7e3a6cb315183b/path/src/transform.rs#L286-L290).

RazrFalcon: Ok, you call transform invert as well... Let's leave it as is then.

src/lib.rs
```diff
@@ -658,15 +674,25 @@ impl FromData for TableRecord {
...
 
 #[cfg(feature = "variable-fonts")]
-#[derive(Clone, Default)]
+#[derive(Clone)]
 struct VarCoords {
     data: [NormalizedCoordinate; MAX_VAR_COORDS],
     len: u8,
 }
 
+#[cfg(feature = "variable-fonts")]
```
RazrFalcon, line 686: Why not derive?

src/lib.rs
```diff
@@ -2218,6 +2254,16 @@ impl<'a> Face<'a> {
         Some(())
     }
 
+    /// Sets the foreground color of the face. Is used by the COLRv1 table.
+    pub fn set_foreground_color(&mut self, color: RgbaColor) {
```
RazrFalcon, line 2258: Hm... I would prefer passing it with each colr1 "outline" call. Would it be possible?

tests/tables/colr.rs
```diff
@@ -56,49 +56,474 @@ fn basic() {
...
+    #[test]
+    fn translate() {
+        let mut face = Face::parse(COLR1_VARIABLE, 0).unwrap();
+        face.set_variation(Tag::from_bytes(b"TLDX"), 100.0);
+        let mut vec_painter = VecPainter(vec![]);
+        face.paint_color_glyph(GlyphId(114), 0, &mut vec_painter);
+
+        assert!(vec_painter.0.contains(&Translate(99.975586, 100.0)));
+    }
+}
```
RazrFalcon, line 529: New line.

src/tables/colr.rs
```diff
@@ -14,6 +23,139 @@ struct BaseGlyphRecord {
...
+    fn default() -> Self {
+        Self {
+            data: &[],
+            records: LazyArray32::default(),
+        }
+    }
+}
+
+impl<'a> ClipList<'a> {
+    #[inline]
```
RazrFalcon, line 102: Isn't this too big for an inline?

src/tables/colr.rs
```diff
@@ -14,6 +23,139 @@ struct BaseGlyphRecord {
...
+
+/// A [clip list](
+/// https://learn.microsoft.com/en-us/typography/opentype/spec/colr#baseglyphlist-layerlist-and-cliplist).
+#[derive(Clone, Copy, Debug)]
+struct ClipList<'a> {
+    data: &'a [u8],
+    records: LazyArray32<'a, ClipRecord>,
+}
+
+impl Default for ClipList<'_> {
```
RazrFalcon, line 92: Can we derive here?

src/tables/colr.rs
```diff
@@ -47,16 +189,520 @@ impl FromData for LayerRecord {
...
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("LinearGradient")
+            .field("x0", &self.x0)
+            .field("y0", &self.y0)
+            .field("x1", &self.x1)
+            .field("y1", &self.y1)
+            .field("x2", &self.x2)
+            .field("y2", &self.y2)
+            .field("extend", &self.extend)
+            // .field("stops", &self.stops(0))
```
RazrFalcon, line 396: ?

src/tables/colr.rs
```diff
@@ -47,16 +189,520 @@ impl FromData for LayerRecord {
...
+            .finish()
+    }
+}
+
+impl<'a> LinearGradient<'a> {
+    /// Returns an iterator over the stops of the linear gradient.
```
RazrFalcon, line 402: Should we mention that those stops should be sorted by the caller?

src/tables/colr.rs
```diff
@@ -47,16 +189,520 @@ impl FromData for LayerRecord {
...
+            #[cfg(feature = "variable-fonts")]
+            coords,
+        }
+    }
+}
+
+/// A [radial gradient](https://learn.microsoft.com/en-us/typography/opentype/spec/colr#formats-6-and-7-paintradialgradient-paintvarradialgradient)
+#[derive(Clone)]
+pub struct RadialGradient<'a> {
+    /// The `x0` value of the radial gradient.
```
RazrFalcon, line 423: I think `of the radial gradient` could be removed.  It's unnecessary.

src/tables/colr.rs
```diff
@@ -47,16 +189,520 @@ impl FromData for LayerRecord {
...
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("RadialGradient")
+            .field("x0", &self.x0)
+            .field("y0", &self.y0)
+            .field("r0", &self.r0)
+            .field("r1", &self.r1)
+            .field("x1", &self.x1)
+            .field("y1", &self.y1)
+            .field("extend", &self.extend)
+            // .field("stops", &self.stops(0))
```
RazrFalcon, line 452: Same missing code.

src/tables/colr.rs
```diff
@@ -14,6 +23,139 @@ struct BaseGlyphRecord {
     num_layers: u16,
 }
 
+/// A [ClipBox](https://learn.microsoft.com/en-us/typography/opentype/spec/colr#baseglyphlist-layerlist-and-cliplist).
```
RazrFalcon, line 26: We already have a `f32` rect called `BBox` in lib.rs.
How about we rename it to `RectF` and then `ClipBox` will be just a typedef?

src/tables/colr.rs
```diff
@@ -14,6 +23,139 @@ struct BaseGlyphRecord {
...
+    /// A paint with a sweep gradient.
+    SweepGradient(SweepGradient<'a>),
+}
+
+/// A [clip record](
+/// https://learn.microsoft.com/en-us/typography/opentype/spec/colr#baseglyphlist-layerlist-and-cliplist).
+#[derive(Clone, Copy, Debug)]
+struct ClipRecord {
+    /// The first glyph ID for the range covered by this record.
+    pub start_glyph_id: GlyphId,
```
RazrFalcon, line 57: No need to be `pub`?

src/tables/colr.rs
```diff
@@ -91,43 +751,1153 @@ impl<'a> Table<'a> {
...
+            let count = s.read::<u32>()?;
+            table.layer_paint_offsets = s.read_array32::<Offset32>(count)?;
+        }
+
+        if let Some(offset) = clip_list_offset {
+            table.clip_list_offsets_offset = offset;
+            let clip_data = data.get(offset.to_usize()..)?;
+            let mut s = Stream::new(clip_data);
+            // Format
+            s.read::<u8>()?;
```
RazrFalcon, line 802: We have `skip` method for this. And put the comment after `;`.

src/tables/colr.rs
```diff
@@ -91,43 +751,1153 @@ impl<'a> Table<'a> {
...
+            let count = s.read::<u32>()?;
+            table.clip_list = ClipList {
+                data: clip_data,
+                records: s.read_array32::<ClipRecord>(count)?,
+            };
+        }
+
+        #[cfg(feature = "variable-fonts")]
```
RazrFalcon, line 810: Have you tried parsing a variable colrv1 font without this feature? Just to make sure.

src/tables/colr.rs
```diff
@@ -91,43 +751,1153 @@ impl<'a> Table<'a> {
...
+            }
+        }
+
+        Some(table)
     }
 
-    fn get(&self, glyph_id: GlyphId) -> Option<BaseGlyphRecord> {
+    /// Returns `true` if the current table has version 0.
+    ///
+    /// A simple table can only emit `outline_glyph`, `paint`
```
RazrFalcon, line 834: _and_ instead of _,_ ?

src/tables/colr.rs
```diff
@@ -91,43 +751,1153 @@ impl<'a> Table<'a> {
...
+
+#[cfg(feature = "variable-fonts")]
+#[derive(Clone, Copy, Debug, Default)]
+struct VariationData<'a> {
+    variation_store: Option<ItemVariationStore<'a>>,
+    delta_map: Option<DeltaSetIndexMap<'a>>,
+}
+
+#[cfg(feature = "variable-fonts")]
+impl VariationData<'_> {
```
RazrFalcon, line 1874: No way we can avoid const generics?
I haven't familiarized myself with them yet. This is not about MSRV, but more about maintenance.

src/tables/cpal.rs
```diff
@@ -20,7 +20,7 @@ impl<'a> Table<'a> {
         let mut s = Stream::new(data);
 
         let version = s.read::<u16>()?;
-        if version != 0 {
+        if version > 1 {
```
RazrFalcon, line 23: v0 and v1 are the same?

examples/font2svg.rs
```diff
@@ -1,10 +1,13 @@
...
 use std::path::PathBuf;
+use base64::engine::general_purpose::STANDARD;
 
 use ttf_parser as ttf;
+use ttf_parser::colr::{ClipBox, Paint};
+use ttf_parser::Transform;
 
 const FONT_SIZE: f64 = 128.0;
-const COLUMNS: u32 = 100;
+const COLUMNS: u32 = 20;
```
RazrFalcon, line 10: Is it necessary? I don't mind, but it would look weird for normal fonts.

RazrFalcon:
Once again a superb job! The code looks like if I wrote it myself, back in those days when I had time...

I guess it was a good introduction into TrueType insanity, especially since you had to deal with variable fonts as well.

And with this change I guess `tff-parser` become one of the most feature-rich TrueType libraries out there. There is not much left to do after that.

Also update the readme. We currently have `COLR table ~ (only v0)`

Have you thought about rasterizing COLRv1 via `tiny-skia` in `font2svg`? Maybe that would be a better choice since SVG is a bit limited for this case.
And we would probably have to do the same in `resvg` anyway.

Cargo.toml
LaurenzV: Yep... It's a dev dependency after all, so it should be fine. 😄

src/lib.rs
LaurenzV: It tells me it's not implemented for `[NormalizedCoordinate; 64]`... Looks like it only works up to a certain number.

src/lib.rs
LaurenzV: You meab `colr::paint`? I think it should be doable, the only reason I changed it is because I thought it might be nice to align it's behaviour to how we deal with variable fonts.

src/tables/colr.rs
LaurenzV: I guess you might be right...

src/tables/colr.rs
LaurenzV: Seems to work.

src/tables/colr.rs
LaurenzV: Well, the problem is right now you need to pass the variation coordinates directly to the `stops` function... But I gues I can just add empty coordinates for the debug function.

src/tables/colr.rs
LaurenzV: Good idea.

src/tables/colr.rs
LaurenzV: Makes sense. I'll also remove it for the others.

src/tables/colr.rs
LaurenzV: Sounds good.

src/tables/colr.rs
LaurenzV: Nope, should there be a need to? It's only needed in the `colr` file.

src/tables/colr.rs
LaurenzV: Yep, just tried it with `font2svg` and then the variable parts are missing.

src/tables/colr.rs
LaurenzV: Sure you can avoid it, but then you have to duplicate the function of the code all across the place... And since we want to be `no-std` this seems to be the best option.

src/tables/cpal.rs
LaurenzV: The first part of the table should be the same. That's why we have an early return in the function if version is 0.

examples/font2svg.rs
LaurenzV: No sure, I can revert it, it was just so I can inspect it better.

LaurenzV:
> I guess it was a good introduction into TrueType insanity, especially since you had to deal with variable fonts as well.

It was indeed. 😄 It's crazy how many different features there are....

> Have you thought about rasterizing COLRv1 via tiny-skia in font2svg? Maybe that would be a better choice since SVG is a bit limited for this case. And we would probably have to do the same in resvg anyway.

Yeah, this would in theory be the next step. The main blocker for this is that we don't have sweep gradients in `tiny-skia`, and I'm honestly not sure if I really want to also look into how skia works... I think dealing with OpenType is already more than enough for me. 😄 

For now, my basic idea was to port the font2svg part into `usvg`, so that `usvg` basically has a best-effort implementation for flattening text. We should remark that it is only best-effort, and that if the user wants accurate COLRv1, they will have to convert the layouted glyphs themselves. And once we have sweep gradients in `tiny-skia`, we basically do the same in `resvg`.  But as I said, not sure when that will happen, for now I will just try to get the best-effort version into `usvg`. It probably works correctly for more than 95% of the emojis, anyway.

tests/tables/colr.rs
LaurenzV: There should be a newline now.. I'm not sure why it doesn't show up.

LaurenzV:
Btw it looks like `ttf-parser` without the `std` feature is broken? Try running `cargo run --no-default-features --example font2svg --  some_font.ttf  out.svg`, it seems to complain about the `parse` function. It's unrelated to this PR though, it doesn't work on `main` either. Just so you know.

src/lib.rs
LaurenzV: I added it to `paint_colr_glyph` instead.

tests/tables/colr.rs
LaurenzV: image
At least in my editor it shows up now.

src/tables/colr.rs
LaurenzV: Added empty ones.

src/lib.rs
RazrFalcon: Got it. Probably limited to hardcoded value.

src/tables/colr.rs
RazrFalcon: As long as it compiles I'm fine. I'm not sure why Rust doesn't warn about pub fields in private structs.

src/tables/colr.rs
RazrFalcon: Ok then.

src/tables/cpal.rs
RazrFalcon: Do we support v1? Does it even have a useful info for us?

RazrFalcon:
Got it. I will look into adding sweep gradients to tiny-skia somewhere this month and will try to implement a COLRv1 rasterizer as well.

Examples are not intended to be running without default features. It's fine.

Ready to merge?

LaurenzV:
> Got it. I will look into adding sweep gradients to tiny-skia somewhere this month and will try to implement a COLRv1 rasterizer as well.

Okay, good luck with that, feel free to ping me if there is a test case you can't get to render properly.

> Ready to merge?

From my side yes!

LaurenzV:
@RazrFalcon Should I port the "best-effort" solution to usvg? Or do you prefer if `usvg` only supports COLRv0, and COLRv1 always needs to be implemented on the on the caller's side?

RazrFalcon:
Sure. Something would be better then nothing.

LaurenzV:
I'll wait for the next `rustybuzz` release then. Somehow I'm not able to patch the ttf-parser dependencyp properly on the `resvg` side...

RazrFalcon:
Strange. Will bump `rustybuzz` then.

You probably have to path `fontdb` as well.

RazrFalcon:
@LaurenzV `rustybuzz` fails after ttf-parser update. https://github.com/RazrFalcon/rustybuzz/actions/runs/9034679635/job/24827768167

I guess something got changed in the variable fonts support.

LaurenzV:
Oof... Okay, I will try to look into it. Sorry, maybe I should've tested the integration with `rustybuzz` before. 😅

RazrFalcon:
Bumped `ttf-parser` in `fontdb` and `rustybuzz`.

LaurenzV:
@RazrFalcon How hard do you think would it be to add sweep gradients to tiny-skia for someone who has never looked at the Skia code base? 😅 If you think it's manageable maybe I'll take a look.

RazrFalcon:
I think it would be a waster of time. You would have to know how Skia works and what changes I had to do to adapt it to Rust and tiny-skia. Unlike rustybuzz, tiny-skia is very different to Skia.

I will try to look into it soon™.

PR #147, opened 2024-05-10
Fix bug in delta set length calculation

LaurenzV:
Sorry about that, now it should work fine. I got a bit confused about how this is calculated.


LaurenzV:
Should I maybe used `checked_add` and `checked_mul`, too?

RazrFalcon:
No worries. Stuff happens.

As for `checked_add`, I think it's fine for now.

PR #148, opened 2024-05-23
Fix check in parse_charset

LaurenzV:
So, I'm not entirely sure if this is how it's supposed to be resolved. I was using `ttf-parser` as a fuzzer for my subsetter, and realized that `parse_charset` has this check to return `None` when there is only one glyph (i.e. the .notdef glyph) in the font. This means that it will fail to parse the cff table in that case. However, I would argue that a font with just the .notdef glyph should be perfectly fine in theory (even though in practice this obviously doesn't make a lot of sense), because the `.notdef` glyph can have outlines as well.

### DIFF
--- a/src/tables/cff/cff1.rs
+++ b/src/tables/cff/cff1.rs
@@ -892,7 +892,7 @@ impl<'a> Table<'a> {
             Some(charset_id::EXPERT_SUBSET) => Charset::ExpertSubset,
             Some(offset) => {
                 let mut s = Stream::new_at(data, offset)?;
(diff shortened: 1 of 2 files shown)


RazrFalcon:
I'm not sure. If we have just one glyph in a font then charset will be empty. But I guess this way we would have an empty charset and not `None`. Very confusing.
Ok, I guess this change is correct.

RazrFalcon:
Can you replace the check with `number_of_glyphs == 0` for readability?

RazrFalcon:
Then `parse_charset` should accept `NonZeroU16` and not `u16`.

PR #151, opened 2024-06-14
Split vvar and hvar

LaurenzV:
Not sure why they were together before, as they have pretty different entries (maybe they added new ones?). But anyway, I had to split them up so that I can add support for variations in `y_origin`, and while I was at I also added the other fields that seem to be in the spec.

See also
https://learn.microsoft.com/en-us/typography/opentype/spec/vvar
https://learn.microsoft.com/en-us/typography/opentype/spec/hvar

### DIFF
--- a/benches/methods_perf_x1000.rs
+++ b/benches/methods_perf_x1000.rs
@@ -80,7 +80,7 @@ fn glyph_hor_advance(bencher: &mut bencher::Bencher) {
     })
 }
 
-fn glyph_left_side_bearing(bencher: &mut bencher::Bencher) {
+fn glyph_hor_side_bearing(bencher: &mut bencher::Bencher) {
     let font_data = std::fs::read("fonts/SourceSansPro-Regular.ttf").unwrap();
     let face = ttf::Face::parse(&font_data, 0).unwrap();
     bencher.iter(|| {
@@ -100,6 +100,6 @@ bencher::benchmark_group!(
     subscript_metrics,
     x_height,
     glyph_hor_advance,
-    glyph_left_side_bearing
+    glyph_hor_side_bearing
 );
 bencher::benchmark_main!(perf);
--- a/benches/methods_perf_x1000.rs
+++ b/benches/methods_perf_x1000.rs
@@ -85,7 +85,7 @@ fn glyph_left_side_bearing(bencher: &mut bencher::Bencher) {
     let face = ttf::Face::parse(&font_data, 0).unwrap();
     bencher.iter(|| {
         for _ in 0..1000 {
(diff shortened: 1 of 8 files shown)


RazrFalcon:
I do not remember why they weren't separate. Probably because `hmtx`/`vmtx` aren't.
Afaiu, the only difference is presence of `y_origin` in `VVAR`.

Also, I would prefer to keep public API the same.

LaurenzV:
If we keep them the same, how should we distinguish them from right side bearing / bottom side bearing then? Right now I haven't added a public API for them because we don't need them, but we might in the future, no?

RazrFalcon:
No, I don't mind the HVAR/VVAR split. I just want to keep the old naming for `glyph_hor_side_bearing` and friends.

LaurenzV:
So what I mean is: The `hvar` table can have two pieces of information: the _left side bearing_ offset and the _right side bearing_ offset. See also the table here:
image

So if we just keep the API as "glyph_hor_side_bearing", then it is not clear whether it refers to the left side bearing or the right side bearing, it could be either since both, left and right are "horizontal"... The same for applies to `glyph_ver_side_bearing", which could be either the top side bearing or the bottom side bearing.

I mean if you want I will still change it back, but I just wanted to let you know.

RazrFalcon:
Honestly, I'm not sure how "right side bearing" offset should work to begin with. We do not have right/bottom side bearing to begin with.
Left/top works by adding a variation offset to the base value, but there is not base value for right/bottom. 
Am I missing something?

In any case, I suggest keeping public API unchanged.

LaurenzV:
> Honestly, I'm not sure how "right side bearing" offset should work to begin with. We do not have right/bottom side bearing to begin with.

Not sure, all I know is that it's in the spec. :p

> In any case, I suggest keeping public API unchanged.

Got it, I'll change it back.

LaurenzV:
Okay, so I changed back the function names in `Face`. I'm not sure if you also wanted me to change them in `vvar` and `hvar`, if yes then I'm not sure how to deal with the naming conflict...

benches/methods_perf_x1000.rs
```diff
@@ -80,7 +80,7 @@ fn glyph_hor_advance(bencher: &mut bencher::Bencher) {
     })
 }
 
-fn glyph_hor_side_bearing(bencher: &mut bencher::Bencher) {
+fn glyph_left_side_bearing(bencher: &mut bencher::Bencher) {
```
RazrFalcon, line 83: Keep old name.

RazrFalcon:
All good. Fix benches and we're done.

PR #152, opened 2024-06-15
Fix treating null offsets as error in GSUB

LaurenzV:
So I was facing problems with a new font added in harfbuzz, and turns out that the reason was that a specific GSUB lookup failed to parse properly. After looking at the spec, I noticed the following:
image

Offsets cann be NULL, which wasn't checked for in our implementation. This is the least invasive approach I could think of (requires no changes in rustybuzz), and the test case passes now as well, so I hope it's okay.

### DIFF
--- a/src/ggg/chained_context.rs
+++ b/src/ggg/chained_context.rs
@@ -1,5 +1,5 @@
 use super::{ClassDefinition, Coverage, SequenceLookupRecord};
-use crate::parser::{FromSlice, LazyArray16, LazyOffsetArray16, Stream};
+use crate::parser::{FromSlice, LazyArray16, LazyOffsetArray16, Offset, Offset16, Stream};
 
 /// A [Chained Contextual Lookup Subtable](
 /// https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#chseqctxt1).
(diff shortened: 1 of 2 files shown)


src/ggg/chained_context.rs
```diff
@@ -44,9 +44,18 @@ impl<'a> ChainedContextLookup<'a> {
             }
             2 => {
                 let coverage = Coverage::parse(s.read_at_offset16(data)?)?;
-                let backtrack_classes = ClassDefinition::parse(s.read_at_offset16(data)?)?;
-                let input_classes = ClassDefinition::parse(s.read_at_offset16(data)?)?;
-                let lookahead_classes = ClassDefinition::parse(s.read_at_offset16(data)?)?;
+                let backtrack_classes = match s.read::<Offset16>()?.to_usize() {
```
RazrFalcon, line 47: Will `Option<Offset16>` look nicer?

RazrFalcon, line 47: Will `Option<Offset16>` look nicer?

RazrFalcon:
Looks good. I guess I've missed this edge case.

src/ggg/chained_context.rs
LaurenzV: I don't get it, like this?

```rs
let backtrack_classes = match s.read::<Option<Offset16>>()?.to_usize() {
    Some(0) => ClassDefinition::Empty,
    Some(offset) => ClassDefinition::parse(data.get(offset..)?)?,
    None => ??
};
```

RazrFalcon: Like this:

```rust
let backtrack_classes = match s.read::<Option<Offset16>>()? {
    Some(offset) => ClassDefinition::parse(data.get(offset.to_usize()..)?)?,
    None => ClassDefinition::Empty,
};
```

LaurenzV: Ah! I didn't know that `Option<Offset16>` has its own implementation for parse. 😄

PR #154, opened 2024-06-17
Add API for getting the clip box of a COLR glyph

LaurenzV:
Needed for 6.0.0.


src/tables/colr.rs
```diff
@@ -866,6 +866,21 @@ impl<'a> Table<'a> {
         self.get_v1(glyph_id).is_some() || self.get_v0(glyph_id).is_some()
     }
 
+    /// Return the clip box for a glyph.
```
RazrFalcon, line 869: "Return**s**" is a preferred Rust convention.

PR #155, opened 2024-06-18
Simplify COLR API for transforms

LaurenzV:
Will make it easier to port code from harfbuzz to rustybuzz. I hope I did the `libm` thing correctly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ default = ["std", "opentype-layout", "apple-layout", "variable-fonts", "glyph-na
 # Enables the use of the standard library. Deactivate this and activate the no-std
 # feature to compile for targets that don't have std.
 std = []
-no-std = ["libm"]
+no-std-float = ["libm"]
 # Enables variable fonts support. Increases binary size almost twice.
 # Includes avar, CFF2, fvar, gvar, HVAR, MVAR and VVAR tables.
 variable-fonts = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,9 +12,15 @@ readme = "README.md"
 edition = "2018"
 exclude = ["benches/**"]
 
+[dependencies]
+libm = { version = "0.2.8", optional = true }
+
 [features]
 default = ["std", "opentype-layout", "apple-layout", "variable-fonts", "glyph-names"]
+# Enables the use of the standard library. Deactivate this and activate the no-std
+# feature to compile for targets that don't have std.
 std = []
+no-std = ["libm"]
 # Enables variable fonts support. Increases binary size almost twice.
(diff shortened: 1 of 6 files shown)


LaurenzV:
Hmm... yeah some advice on how to deal with the feature thingy would be appreciated. 😄

Cargo.toml
```diff
@@ -12,9 +12,15 @@ readme = "README.md"
...
 
+[dependencies]
+libm = { version = "0.2.8", optional = true }
+
 [features]
 default = ["std", "opentype-layout", "apple-layout", "variable-fonts", "glyph-names"]
+# Enables the use of the standard library. Deactivate this and activate the no-std
+# feature to compile for targets that don't have std.
 std = []
+no-std = ["libm"]
```
RazrFalcon, line 23: I would suggest using `no-std-float`, just like in `tiny-skia`.

src/lib.rs
```diff
@@ -432,6 +432,44 @@ impl Transform {
         Transform::new(1.0, 0.0, 0.0, 1.0, tx, ty)
     }
 
+    /// Creates a new rotation transform.
+    #[inline]
+    pub fn new_rotate(angle: f32) -> Self {
+        #[cfg(feature = "no-std")]
```
RazrFalcon, line 438: I would suggest going the `tine-skia` route by using [a trait](https://github.com/RazrFalcon/tiny-skia/blob/784569522f728cc70232cbb88abd7330571b51dd/path/src/scalar.rs#L78-L125).

And then 
```rust
#[cfg(all(not(feature = "std"), feature = "no-std-float"))]
use crate::NoStdFloat;
```

where needed.

RazrFalcon: You can put the new trait somewhere in `lib.rs`

LaurenzV:
I'll mark this as draft since I also want to test it in `rustybuzz` first, which might take a while.

LaurenzV:
Is it okay that when building with no default features and `variable-fonts`, you must also add `no-std-float` (or std)? Kinda annoying, but not sure if there is a better way...

RazrFalcon:
Yes, same with `tiny-skia`. `--no-default-features` implies adding `--feature libm`. No better way to handle this to my knowledge.

LaurenzV:
Okay, should be good now.

Cargo.toml
```diff
@@ -12,9 +12,15 @@ readme = "README.md"
 edition = "2018"
 exclude = ["benches/**"]
 
+[dependencies]
+libm = { version = "0.2.8", optional = true }
+
 [features]
 default = ["std", "opentype-layout", "apple-layout", "variable-fonts", "glyph-names"]
+# Enables the use of the standard library. Deactivate this and activate the no-std
```
RazrFalcon, line 20: `no-std-float`, not `no-std`

PR #156, opened 2024-06-21
Add support for getting phantom points

LaurenzV:
So, it seems that at least for no variations, the calculated phantom points seem to match the ones from harfbuzz.

However, I'm having somewhat of a hard time figuring out how to best apply the variations to it. What harfbuzz seems to be doing is to just append the points to the glyph points and then process the variations on the vector of glyphs. I tried into looking how we can best implement this, but the big problem seems that the `ttf-parser` implementation works quite differently since we are using iterators, so we cannot just append the phantom points and treat them as regular points. 

And the `parse_variation_data` function from `gvar` requires a bunch of iterators, so I'm not sure what the best way is to call them to just calculate the variations for the phantom points. 

Any help would be appreciated.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2341,6 +2341,38 @@ impl<'a> Face<'a> {
         }
     }
 
+    /// Return an array containing the phantom points of a glyph. The first element
+    /// of a tuple contains the x coordinate, the second the y coordinate. They are
+    /// arranged in the order `PHANTOM_LEFT`, `PHANTOM_RIGHT`, `PHANTOM_TOP`,
+    /// `PHANTOM_BOT`.
+    #[inline]
+    pub fn phantom_points(&self, glyph_id: GlyphId) -> Option<[(i16, i16); 4]> {
+        let bbox = self.tables.glyf?.bbox(glyph_id).unwrap_or(Rect::zero());
+
+        let lsb = self.glyph_hor_side_bearing(glyph_id);
+        let mut phantoms = [(0, 0); 4];
+        let h_delta = if let Some(lsb) = lsb {
+            bbox.x_min - lsb
+        } else {
+            0
+        };
+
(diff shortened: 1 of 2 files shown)


RazrFalcon:
I will try to find time to look into it.

RazrFalcon:
I have added basic phantom points support. That fixed existing `rustybuzz` issues, but it's not a proper implementation.
But since `harfbuzz` doesn't seem to test other cases we ignore them as well. I'm talking specifically about the code you wrote in this PR.

As for the fix, you will notice that it's pretty cryptic and very different to HB implementation. This is by design. `glyf`/`gvar` implementations in `ttf-parser` and `harfbuzz` are completely different.

One thing to note is that `gvar` and `CFF` are two most insane TrueType tables. It took me a month to implement each of them.
As for phantom points, the idea is following (assuming I understand it correctly):
- `gvar` stores deltas that apply to individual `glyf` points
- at the end of the deltas list for each glyph are phantom points; basically just 4 more deltas
- to access them we have to parse/skip regular deltas, but their amount isn't stored anywhere, therefore we have to parse the number of points in `glyf` first
- after we got those 4 deltas we have to apply them somehow to glyph's bbox and other metrics, but I have no idea how and when

While the [`gvar` spec](https://learn.microsoft.com/en-us/typography/opentype/spec/gvar) is pretty verbose and detailed, it's mostly pointless gibberish. I have no idea what they talking about. Please, just show me the code!

We do not support `USE_MY_METRICS` as well, which apparently harfbuzz do not test as well... Therefore in terms of glyf/gvar we're still missing proper phantom points support, matching points support, instructions support (think glyph adjusting assembly) and `USE_MY_METRICS`.

PS: we should probably check if `read-fonts` support them

LaurenzV:
> PS: we should probably check if read-fonts support them

They most definitely do support it in some form considering they support hinting for which it is required, but I would have to check to what extent / what API they provide.

PR #160, opened 2024-06-27
clippy run

asibahi:
Went over the project with clippy.

One thing I don't understand is the errors in the benches folder. What's the deal with the undeclared `bencher` module? How do.I deal with it?


src/tables/glyf.rs
```diff
@@ -618,18 +618,21 @@ impl<'a> Table<'a> {
...
-            let glyph_points = parse_simple_outline(s.tail()?, number_of_contours)?;
-            Some(glyph_points.points_left)
-        } else if number_of_contours < 0 {
-            // Composite glyph.
-            let components = CompositeGlyphIter::new(s.tail()?);
-            Some(components.clone().count() as u16)
-        } else {
-            // An empty glyph.
-            None
+        use core::cmp::Ordering::*;
```
RazrFalcon, line 621: Keep the original code here.

RazrFalcon:
Thanks, but for the future I do not accept clipping and formatting fixes.

asibahi:
> Thanks, but for the future I do not accept clipping and formatting fixes.

Noted, and thank you.

I am wondering about the the `bencher` thing that does not allow me to run `cargo bench` . Are there instructions to resolve that that I missed?

Cargo.toml
```diff
@@ -46,3 +46,4 @@ base64 = "0.22.1"
 pico-args = "0.5"
 tiny-skia-path = "0.11.4"
 xmlwriter = "0.1"
+bencher = "0.1"
```
RazrFalcon, line 49: This one isn't needed.

asibahi: Yes but it makes the IDE a lot calmer.

RazrFalcon: Then it's an IDE bug. `ttf-parser/benches` has its own `Cargo.toml`. It's a completely separate entity.

RazrFalcon:
Looks like `data.extend([2; 100]);` is a fairly new feature. Let's revert it as well.

PR #162, opened 2024-06-30
Use core maths

LaurenzV:
So what I did here is that I always include the `core_maths` dependency and I removed `no-std-float`, so that in order to enable `no_std` mode, you can simply disable all default features and don't need to additionally enable a separate feature. The reason for this is that since features are additive, it's apparently [bad practice to have an explicity `no_std` feature.](https://www.reddit.com/r/rust/comments/f2fsi8/comment/fhcji58). Because if one crate depends on `std` of `ttf-parser` and another on `no-std-float`, then both features would be activated if a crates uses both of them as a dependency.

If you want I can undo the last part, though.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,8 +24,8 @@ jobs:
         toolchain: ${{ matrix.rust }}
         override: true
 
-    - name: Build with no-std-float
-      run: cargo build --no-default-features --features no-std-float
+    - name: Build with no default features
+      run: cargo build --no-default-features
 
     - name: Build with std
       run: cargo build --no-default-features --features std
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,7 +31,7 @@ jobs:
       run: cargo build --no-default-features --features std
 
     - name: Build with variable-fonts
-      run: cargo build --no-default-features --features variable-fonts,no-std-float
+      run: cargo build --no-default-features --features variable-fonts
 
     - name: Build with all features
       run: cargo build --all-features
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,14 +13,12 @@ edition = "2018"
 exclude = ["benches/**"]
 
 [dependencies]
-libm = { version = "0.2.8", optional = true }
+core_maths = { version = "0.1.0"}
 
 [features]
 default = ["std", "opentype-layout", "apple-layout", "variable-fonts", "glyph-names"]
-# Enables the use of the standard library. Deactivate this and activate the no-std-float
-# feature to compile for targets that don't have std.
+# Enables the use of the standard library.
 std = []
-no-std-float = ["libm"]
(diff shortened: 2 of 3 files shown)


RazrFalcon:
Yep, looks good to me.

alexheretic:
This means _ttf-parser_ now depends on _core_maths_ & _libm_ always. Part of the point of having a "libm" feature (or equivalent) is to opt into it when you need it for no-std compilation and to avoid additional dependencies otherwise for the (quite common) "std" case.

Previously _ttf-parser_ had no dependencies.

```
ttf-parser v0.23.0
└── core_maths v0.1.0
    └── libm v0.2.8
```

alexheretic:
Yep it's a tradeoff, just seems worth considering for this crate as it is so lean on dependencies.

Having the extra dependencies using "std" doesn't seem to affect the size of binary afaics. But there is a minor regression in clean compile times.

### 0.22
```
$ hyperfine --runs 6 --prepare 'rm -rf ./target' 'cargo build --target-dir target'
Benchmark 1: cargo build --target-dir target
  Time (mean ± σ):      1.205 s ±  0.009 s    [User: 1.489 s, System: 0.172 s]
  Range (min … max):    1.188 s …  1.213 s    6 runs
```

### master (0.23)
```
$ hyperfine --runs 6 --prepare 'rm -rf ./target' 'cargo build --target-dir target'
Benchmark 1: cargo build --target-dir target
  Time (mean ± σ):      1.439 s ±  0.030 s    [User: 1.771 s, System: 0.212 s]
  Range (min … max):    1.411 s …  1.497 s    6 runs
```

alexheretic:
I think the previous approach, having a feature, was mildly better.

> And if you have two crates, one of which depends on std ttf-parser and one depends on no_std ttf-parser - then you will get core_maths/libm anyway.

Yes but as I said before this is an edge case where the downside is the same downside now applied to everyone.

Using a feature, a crate downstream of _ttf-parser_ can still also support no-std with a "libm" feature and use that to activate _ttf-parser_'s feature. E.g. 
```toml
[features]
libm = ["dep:libm", "ttf-parser/libm"]
``` 
So there isn't a good reason that the dependency tree will inevitably have no-std _ttf-parser_ in it anyway as there is a pattern to activating it when necessary.

I guess the argument for the ergonomics is that no-std users already have to set `default-features = false` so it isn't much harder to add `features = ["libm"]`.

So imo having the feature is generally the best approach because it minimises dependencies. I've used this approach myself too (e.g. [ab_glyph_rasterizer](https://github.com/alexheretic/ab-glyph/blob/main/rasterizer/README.md#no_std)) and I don't think it causes a problem for no-std usage.

But overall, I can live with either way. So I've probably banged on about this enough now! :sweat_smile:

RazrFalcon:
Would making `core_maths` an optional dependency will be a good enough solution? `libm` requires a bit more work.

PR #165, opened 2024-07-21
Parse names for Macintosh platform ID + Roman

hut36:


RazrFalcon:
1. The idea is that language names are parsed by the caller.
2. Your code is incorrect. MacRoman is not UTF-16. You can check how [fontdb](https://github.com/RazrFalcon/fontdb/blob/master/src/lib.rs) handles this.

hut36:
> 1. The idea is that language names are parsed by the caller.
> 
>     2. Your code is incorrect. MacRoman is not UTF-16. You can check how [fontdb](https://github.com/RazrFalcon/fontdb/blob/master/src/lib.rs) handles this.

Thanks for your reply, it really helps! Checked fontdb, and I think it fits my needs.

PR #166, opened 2024-08-12
Add support for the STAT table.

inferiorhumanorgans:
https://learn.microsoft.com/en-us/typography/opentype/spec/stat

This PR adds support for parsing the `STAT` table and its related data types.  It's a pretty simple table but as in #87 the issues of testing and API design deserve some consideration.

Additionally, should this be placed behind a feature?

### DIFF
--- a/examples/font-info.rs
+++ b/examples/font-info.rs
@@ -1,4 +1,4 @@
-use ttf_parser::stat::AxisValue;
+use ttf_parser::stat::{AxisValue, AxisValueTable};
 
 fn main() {
     let args: Vec<_> = std::env::args().collect();
@@ -101,40 +101,74 @@ fn main() {
         }
 
         println!("  Axis Values:");
-        for value in stat.values() {
-            // println!("    {value:?}");
-            match value {
-                AxisValue::Format1(value) => {
-                    let name = face
+        for table in stat.tables() {
+            match table {
+                AxisValueTable::Format1(table) => {
+                    let value_name = face
                         .names()
                         .into_iter()
-                        .filter(|name| name.name_id == value.value_name_id)
+                        .filter(|name| name.name_id == table.value_name_id)
                         .map(|name| name.to_string().unwrap())
                         .collect::<Vec<_>>()
                         .join(", ");
 
+                    let axis_name = &axis_names[table.axis_index as usize];
+                    let value = table.value;
+                    let flags = table.flags;
+
+                    println!("    {axis_name} {value:?}={value_name:?} flags={flags:?}");
+                }
+                AxisValueTable::Format2(table) => {
+                    let value_name = face
+                        .names()
+                        .into_iter()
+                        .filter(|name| name.name_id == table.value_name_id)
+                        .map(|name| name.to_string().unwrap())
+                        .collect::<Vec<_>>()
+                        .join(", ");
+
+                    let axis_name = &axis_names[table.axis_index as usize];
+                    let nominal_value = table.nominal_value;
+                    let min_value = table.range_min_value;
+                    let max_value = table.range_max_value;
+                    let flags = table.flags;
+
+                    println!("    {axis_name} {min_value:?}..{max_value:?}={value_name:?} nominal={nominal_value:?} flags={flags:?}");
+                }
+                AxisValueTable::Format3(table) => {
+                    let value_name = face
+                        .names()
+                        .into_iter()
+                        .filter(|name| name.name_id == table.value_name_id)
+                        .map(|name| name.to_string().unwrap())
+                        .collect::<Vec<_>>()
+                        .join(", ");
+
+                    let axis_name = &axis_names[table.axis_index as usize];
+                    let value = table.value;
+                    let linked_value = table.linked_value;
+                    let flags = table.flags;
+
                     println!(
-                        "    {}={:?}={name:?} flags={:?}",
-                        &axis_names[value.axis_index as usize], value.value, value.flags
+                        "    {axis_name} {value:?}<=>{linked_value:?} = {value_name:?} flags={flags:?}",
                     );
                 }
-                AxisValue::Format2(_) => todo!(),
-                AxisValue::Format3(value) => {
-                    let name = face
+                AxisValueTable::Format4(table) => {
+                    let value_name = face
                         .names()
                         .into_iter()
-                        .filter(|name| name.name_id == value.value_name_id)
+                        .filter(|name| name.name_id == table.value_name_id)
                         .map(|name| name.to_string().unwrap())
                         .collect::<Vec<_>>()
                         .join(", ");
 
-                    println!(
-                        "    {} {:?}<=>{:?} = {name:?} flags={:?}",
-                        &axis_names[value.axis_index as usize],
-                        value.value,
-                        value.linked_value,
-                        value.flags
-                    );
+                    let flags = table.flags;
+
+                    println!("    {value_name:?} flags={flags:?}");
+                    for pair in table.values {
+                        let AxisValue { axis_index, value } = pair;
+                        println!("      {axis_index} = {value:?}")
+                    }
                 }
             }
         }
--- a/examples/font-info.rs
+++ b/examples/font-info.rs
@@ -1,5 +1,3 @@
-use ttf_parser::stat::{AxisValue, AxisValueSubtable};
-
 fn main() {
     let args: Vec<_> = std::env::args().collect();
     if args.len() != 2 {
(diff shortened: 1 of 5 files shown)


src/tables/stat.rs
```diff
@@ -0,0 +1,336 @@
...
+            data,
+            value_lookup_start,
+            value_offsets,
+            fallback_name_id,
+            version,
+        })
+    }
+
+    /// Iterator over the collection of axis value tables.
+    pub fn tables(&self) -> AxisValueTables<'a> {
```
RazrFalcon, line 327: Maybe "subtables"? This is a common terminology.

src/tables/stat.rs
```diff
@@ -0,0 +1,336 @@
...
+
+/// A [Style Attributes Table](https://docs.microsoft.com/en-us/typography/opentype/spec/stat).
+#[derive(Clone, Copy, Debug)]
+pub struct Table<'a> {
+    /// List of axes
+    pub axes: LazyArray16<'a, AxisRecord>,
+    /// Fallback name when everything can be elided.
+    pub fallback_name_id: Option<u16>,
+    /// Version of the style attributes table.
+    pub version: u32,
```
RazrFalcon, line 276: Do we have to expose the version?

examples/font-info.rs
```diff
@@ -84,6 +86,94 @@ fn main() {
         }
     }
 
+    if let Some(stat) = face.tables().stat {
```
RazrFalcon, line 89: Let's put it into a separate functions, like with `print_opentype_layout`.

Also, it's pretty verbose. Maybe there is a way to make the code shorter. Maybe we should expose a more high-level API as well, to avoid direct subtables accesses? Would it be doable?

RazrFalcon:
Thanks! Looks pretty good to me.

yisibl:
It looks like we can update the table in the readme.
https://github.com/RazrFalcon/ttf-parser?tab=readme-ov-file#alternatives

src/tables/stat.rs
inferiorhumanorgans: It depends on how strict the parsing should be.  v1.2 introduced a new axis value table type (format 4).  If it's okay to accept format 4 entries in an older version of the `STAT` table then we can drop the version.

examples/font-info.rs
inferiorhumanorgans: > Maybe we should expose a more high-level API as well, to avoid direct subtables accesses? 

Well anything's possible.  What did you have in mind?

Ignoring the format 4 tables, for a non-variable font I imagine querying by axis tag and getting a single value back makes sense.  For instance Robto Condensed lists a "normal" width in the `OS/2` table, but lists the proper width in the `STAT` table.  So a width check on a non-variable font might look first in `STAT` to see if there's an entry for `wdth` and grab whatever from the `OS/2` table as a fallback.

For a variable font, the query would be by tag and value (based on what's in `fvar`).  E.g. what's the name for `wght` == 500?

Format 4 tables open up the possibility of querying via multiple axes (e.g. what's the stretch at `wght` == 200 and `ital` == 1.0), but that's something I'd prefer to punt on for now.

```rust

enum StyleValue {
  Single(Fixed),
  Range { min: Fixed, max: Fixed, nominal: Fixed },
  Link { value: Fixed, target: Fixed },
}

struct StyleAttribute {
  name: Tag,
  flags: Flags,
  value: StyleValue,
}

fn style_for_axis(tag: Tag) -> StyleAttribute;
fn style_for_axis_and_value(tag: Tag, value: Fixed) -> StyleAttribute;

// Only useful with format 4 tables
fn style_for_values(values: &[(Tag, Fixed)]) -> StyleAttribute;
```

src/tables/stat.rs
RazrFalcon: Checking version during parsing is fine, I just don't see why this field should be public. Version should be handled implicitly by the parser.

examples/font-info.rs
RazrFalcon: I never looked into how the STAT table looks, so I cannot comment for now. I would have to investigate it myself.
I was just curious if the code in this example can be simplified. Right now it's too big for a simple example.

src/tables/stat.rs
```diff
@@ -0,0 +1,336 @@
...
+        let axis_index = s.read::<u16>()?;
+        let value = s.read::<Fixed>()?;
+
+        Some(AxisValue { axis_index, value })
+    }
+}
+
+/// List of axis value tables.
+#[derive(Clone, Debug)]
+pub struct AxisValueTables<'a> {
```
RazrFalcon, line 31: This should become a `Subtables` as well. The only table here should be the STAT itself.

examples/font-info.rs
inferiorhumanorgans: Welp I'm all for exposing a higher level query API.  Agreed that what's in the example is too low level.

RazrFalcon: Let's move the STAT code to a new function for no, as I have suggested earlier. And leave it as is otherwise for now.

inferiorhumanorgans: Ah, by the time I saw this I'd already hacked up a simple query function which I think cleans things up a little bit and creates a foundation to update `is_italic` and `is_bold`.

src/tables/stat.rs
```diff
@@ -332,4 +332,70 @@ impl<'a> Table<'a> {
...
     }
+
+    /// Returns the first matching subtable for a given axis.  If no match value is given the first
+    /// subtable for the axis is returned.  If a match value is given, the first subtable for the
+    /// axis where the value matches is returned.  A value matches if it is equal to the subtable's
+    /// value or contained within the range defined by the subtable.  If no matches are found `None`
+    /// is returned.  Typically a match value is not specified for non-variable fonts as multiple
+    /// subtables for a given axis ought not exist.  For variable fonts a non-`None` match value
+    /// should be specified as multiple records for the variation axes exist.
+    pub fn subtable_for_axis(
```
RazrFalcon, line 343: Ok, so you have decided to create a query method instead. Fine by me.

Can you fix up the docs a bit? Do not use double spaces and make sure the first line is new-line separated. Like:

```rust
    /// Returns the first matching subtable for a given axis.
    /// 
    /// If no match value is given the first subtable for the axis is returned.
    /// 
    /// If a match value is given, the first subtable for the axis
    /// where the value matches is returned.
    /// A value matches if it is equal to the subtable's value
    /// or contained within the range defined by the subtable.
    /// 
    /// If no matches are found `None` is returned. Typically a match value is not specified
    /// for non-variable fonts as multiple subtables for a given axis ought not exist.
    /// For variable fonts a non-`None` match value should be specified
    /// as multiple records for the variation axes exist.
```

examples/font-info.rs
```diff
@@ -201,3 +115,15 @@ fn print_opentype_layout(name: &str, table: &ttf_parser::opentype_layout::Layout
         println!("    {}", feature);
     }
 }
+
+fn print_opentype_style_attributes(table: &ttf_parser::stat::Table) {
```
RazrFalcon, line 119: No need for this anymore since you have extracted the code anyway. We can keep it main.

RazrFalcon:
Ok, 1.51 compatibility is broken, but we're good otherwise.

inferiorhumanorgans:
Ok – I've rolled up a few other minor changes. I've also set up a few functions to abstract over getting the value and flags. If that seems in scope for this PR I can push those as well.

RazrFalcon:
>If that seems in scope for this PR I can push those as well.

Sure, let's take a look.

inferiorhumanorgans:
I'm not sure how far down the rabbit hole it's worth going with providing a higher level API.  What I've exposed so far should be useful for both dynamic (e.g. `name_id`) and non-dynamic (e.g. `contains`) fonts.

For instance:

```rust
fn is_stat_italic() -> bool {
  if let Some(stat) = face.tables().stat {
    if let Some(subtable) = stat.subtable_for_axis(Tag::from_bytes(b"ital"), None) {
      if subtable.contains(Fixed(1.0)) {
        return true;
      }
    }
  }

  return false
}
```

```rust
fn is_bold() -> bool {
  if let Some(stat) = face.tables().stat {
    if let Some(subtable) = stat.subtable_for_axis(Tag::from_bytes(b"wght"), None) {
      if let Some(value) = subtable.value() {
        let named_weight = Weight::from(value.0 as u16);
        if named_weight == Weight::Bold {
          return true;
        }
      }
    }
  }

  return false
}
```

RazrFalcon:
If you think it's good enough I'm fine with merging it.

inferiorhumanorgans:
Works for me.

PR #167, opened 2024-08-14
Add an accessor for the raw Unicode ranges value on the OS/2 table

nwoods-cimpress:
Added `UnicodeRanges::as_raw()` to provide access to the underlying raw value for the `ulUnicodeRangeX` fields on the OS/2 table

### DIFF
--- a/src/tables/os2.rs
+++ b/src/tables/os2.rs
@@ -193,6 +193,11 @@ impl UnicodeRanges {
             false
         }
     }
+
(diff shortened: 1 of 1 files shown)


RazrFalcon:
I would rather use `pub struct UnicodeRanges(pub u128);` (note the second `pub`) instead.

nwoods-cimpress:
Changed!

RazrFalcon:
Thanks!

PR #172, opened 2024-09-07
Initial support for parsing the PANOSE number

inferiorhumanorgans:
This includes a ton of boilerplate (apologies), a few tests, and adds checks for the PANOSE number in is_bold / is_italic / is_monospaced.  

The PANOSE number also encodes a font weight that's distinct from the weight class in the OS/2 table so I've included a function to attempt to map the PANOSE weight to a weight class.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1408,8 +1408,18 @@ impl<'a> Face<'a> {
     /// Checks that face is marked as *Italic*.
     #[inline]
     pub fn is_italic(&self) -> bool {
+        let panose_italic = self
+            .tables
+            .os2
+            .map(|os2| {
+                os2.panose()
+                    .map(|panose| panose.is_italic())
+                    .unwrap_or(false)
+            })
+            .unwrap_or(false);
+
         // A face can have a Normal style and a non-zero italic angle, which also makes it italic.
-        self.style() == Style::Italic || self.italic_angle() != 0.0
+        self.style() == Style::Italic || self.italic_angle() != 0.0 || panose_italic
     }
 
     /// Checks that face is marked as *Bold*.
@@ -1417,7 +1427,19 @@ impl<'a> Face<'a> {
     /// Returns `false` when OS/2 table is not present.
     #[inline]
     pub fn is_bold(&self) -> bool {
-        self.tables.os2.map(|os2| os2.is_bold()).unwrap_or(false)
+        let panose_bold = self
+            .tables
+            .os2
+            .map(|os2| {
+                os2.panose()
+                    .map(|panose| panose.is_italic())
+                    .unwrap_or(false)
+            })
+            .unwrap_or(false);
+
+        let sf_bold = self.tables.os2.map(|os2| os2.is_bold()).unwrap_or(false);
+
+        sf_bold || panose_bold
     }
 
     /// Checks that face is marked as *Oblique*.
@@ -1436,13 +1458,26 @@ impl<'a> Face<'a> {
 
     /// Checks that face is marked as *Monospaced*.
     ///
-    /// Returns `false` when `post` table is not present.
+    /// Returns `false` when neither `post` nor `os2` tables are not present.
     #[inline]
     pub fn is_monospaced(&self) -> bool {
-        self.tables
+        let panose_monospaced = self
+            .tables
+            .os2
+            .map(|os2| {
+                os2.panose()
+                    .map(|panose| panose.is_monospaced())
+                    .unwrap_or(false)
+            })
+            .unwrap_or(false);
+
+        let post_monospaced = self
+            .tables
             .post
             .map(|post| post.is_monospaced)
-            .unwrap_or(false)
+            .unwrap_or(false);
+
+        post_monospaced || panose_monospaced
     }
 
     /// Checks that face is variable.
--- a/src/tables/os2/panose.rs
+++ b/src/tables/os2/panose.rs
@@ -1,0 +1,1295 @@
+//! A [PANOSE classification](
+//! https://learn.microsoft.com/en-us/typography/opentype/spec/os2#panose) implementation.
+//!
+//! The underlying specification is located at: <https://monotype.github.io/panose/pan1.htm>
+
+use crate::{parser::Stream, FromData};
+
+/// The type of diagonal stems and letterform termination used by a font face.
+///
+/// <https://monotype.github.io/panose/pan2.htm#_Toc380547277>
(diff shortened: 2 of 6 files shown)


src/lib.rs
```diff
@@ -1408,16 +1408,38 @@ impl<'a> Face<'a> {
     /// Checks that face is marked as *Italic*.
     #[inline]
     pub fn is_italic(&self) -> bool {
+        let panose_italic = self
```
RazrFalcon, line 1411: Let's make it less functional... We can move a `panose` getter into a private function, similar to `Face::style()`.

src/tables/os2/panose.rs
```diff
@@ -0,0 +1,1295 @@
...
+    NonStraightVertical = 9,
+    NonStraightSingleSerif = 10,
+    NonStraightDoubleSerif = 11,
+}
+
+impl FromData for ArmStyle {
+    const SIZE: usize = 1;
+
+    fn parse(data: &[u8]) -> Option<Self> {
+        if data.is_empty() {
```
RazrFalcon, line 33: No need to check for that. `FromData` caller already did it.

RazrFalcon:
It's definitely verbose... Do we really need all those enums? Maybe we can simply expose:
```rust
struct Panose {
    pub data: [u8; 10]
}
```

plus `is_bold`/etc. getters. And if someone wants to get some super specific/niche values they can do this themselves.

1250 LOC for a deprecated TrueType feature is a lot... Stuff like `aspect_211` is a dead weight, imho.

inferiorhumanorgans:
> It's definitely verbose... Do we really need all those enums? Maybe we can simply expose:
> 
> ```rust
> struct Panose {
>     pub data: [u8; 10]
> }
> ```
> 
> plus `is_bold`/etc. getters. And if someone wants to get some super specific/niche values they can do this themselves.
> 
> 1250 LOC for a deprecated TrueType feature is a lot... Stuff like `aspect_211` is a dead weight, imho.


Yeah it's a lot…

In terms of exposing a 10 byte slice, since it's a fixed length/offset field I think that's more suitable for `RawFaceTables`.

Can the enums be eliminated? Sure. At its most basic this code is currently there to support three questions (is the font face italic/bold/monospaced).  The checks could be collapsed into simply checking the bytes manually, but I think there's a legibility tradeoff.

e.g.

```rust
fn is_italic() -> bool {
  (panose[0] == 2 && (panose[7] >= 9 && panose[7] <= 15)) ||
  (panose[0] == 3 && (panose[7] >= 6 && panose[7] <= 13))
}
```

Unsure what a happy medium would look like.  I doubt that folks are clamoring to build an rtf parser/renderer, but as other use cases get added on (e.g. mapping weight to a weight class, or distinguishing between text / handwriting / symbol/wingdings font) I think having some sort of abstraction makes more sense.

RazrFalcon:
We can keep only enums used by the public API. Aka `is_italic` and friends.
To my understanding, no one uses `panose` anymore. It's an outdated feature. So having a 1000 LOC for something no one will use is an overkill for me.

RazrFalcon:
Any updates on this?

inferiorhumanorgans:
Apologies — I'll have time next week to dig into this.

RazrFalcon:
No problem, just checking in.

Goldziher:
Thank you for this, and I'm sorry it sat so long without a decision — that's on the project, not you. We've taken over maintenance and are working through the backlog.

We're not going to merge this in its current shape, but we would like to land PANOSE, and there's a concrete smaller version we'd merge quickly. Reasoning, so it's not just a verdict:

**Size and permanence.** 1,295 of the 1,383 lines are one file containing 25 public enums and 266 variants, each written twice — once as variants with explicit discriminants, then again as `FromData` match arms restating the same numbers. None are `#[non_exhaustive]` (I checked: zero). At 92.9M downloads every one of those variants is a permanent semver commitment, and PANOSE digit value sets have been extended historically. For comparison, `os2.rs` today exposes four enums — `Weight`, `Width`, `Permissions`, `Style` — and `Weight`/`Width` both carry `to_number()` escape hatches precisely so unnamed values stay reachable. This crate's habit is a few enums plus raw access, not exhaustive modelling of a spec. That's the same point Yevhenii raised in this thread, so it isn't a new bar being applied to you late.

**A bug worth fixing regardless of what happens to this PR.** In `Face::is_bold()`:

```rust
let panose_bold = self.tables.os2
    .map(|os2| os2.panose().map(|panose| panose.is_italic()).unwrap_or(false))
```

That calls `is_italic()`, while `Panose::is_bold()` is defined immediately above and unused here. Any font with an oblique PANOSE letterform would report as **bold**. Nothing in the test suite covers `is_bold`, so it passes green.

**Silent behaviour change.** `is_italic`, `is_bold` and `is_monospaced` each gain a new `|| panose_*` term. Those are long-stable public predicates; existing users would get different answers with no compile error and no test coverage of the new paths. If PANOSE should feed them, that wants to be deliberate and tested — arguably opt-in.

Two smaller things: `Serif` is shared between `LatinText.serif` (digit 1) and `LatinDecorative.serif` (digit 5), which are different value sets in the spec, so it carries `Script = 16` that is only valid on the Decorative side; and `FromData for Panose` returns `None` if *any* of the nine sub-bytes is out of range, so one junk digit discards the whole structure — real-world PANOSE fields are frequently junk or zero-padded, and the crate is lenient elsewhere.

Credit where due: the discriminants I spot-checked against PANOSE 2.0 are correct (`Weight` 0-11, `Proportion` 0-9, `Spacing` 0-3, `Letterform` 0-15 with Oblique at 9-15, `HandForm` 0-13 with Oblique at 6-9), the Any/NoFit short-circuit is handled properly, and the whole thing is clean `no_std` with no allocation. The `os2.rs` → `os2/mod.rs` split is also the right structure and matches `cmap/` and `cff/`. This is careful work; the objection is scope, not craft.

**What we'd merge:** raw PANOSE byte access on `os2::Table`, plus `is_bold`/`is_italic`/`is_monospaced` support, and only the one or two enums those actually need — with the `is_bold` fix and tests covering the changed predicates. That's a small PR and we'd review it promptly. If you'd rather not revisit it after this long, say so and we'll pick it up ourselves with you credited in the changelog.

Leaving this open rather than closing it, so it's your call which way to go.

inferiorhumanorgans:
Sorry, I don't speak AI.  What are you trying to say?

Goldziher:
Sure. Rebased, rewrote and shortened, merged. Thanks for contributing.

PR #174, opened 2024-09-30
Add codepoints_iter() methods for cmap Subtable4/Subtable12

MoSal:
This allows early exit unlike `codepoints()`.

### DIFF
--- a/src/tables/cmap/format12.rs
+++ b/src/tables/cmap/format12.rs
@@ -65,13 +65,16 @@ impl<'a> Subtable12<'a> {
         u16::try_from(id).ok().map(GlyphId)
     }
 
+    /// Iterate over each codepoint defined in this table.
+    pub fn codepoints_iter(&'a self) -> impl Iterator<Item = u32> + 'a {
+        self.groups
+            .into_iter()
+            .flat_map(|group| group.start_char_code..=group.end_char_code)
+    }
+
     /// Calls `f` for each codepoint defined in this table.
-    pub fn codepoints(&self, mut f: impl FnMut(u32)) {
-        for group in self.groups {
-            for code_point in group.start_char_code..=group.end_char_code {
-                f(code_point);
-            }
-        }
+    pub fn codepoints(&self, f: impl FnMut(u32)) {
+        self.codepoints_iter().for_each(f)
     }
 }
 
--- a/src/tables/cmap/format4.rs
+++ b/src/tables/cmap/format4.rs
@@ -100,18 +100,19 @@ impl<'a> Subtable4<'a> {
         None
     }
 
-    /// Calls `f` for each codepoint defined in this table.
-    pub fn codepoints(&self, mut f: impl FnMut(u32)) {
-        for (start, end) in self.start_codes.into_iter().zip(self.end_codes) {
-            // OxFFFF value is special and indicates codes end.
-            if start == end && start == 0xFFFF {
-                break;
-            }
+    /// Iterate over each codepoint defined in this table.
+    pub fn codepoints_iter(&'a self) -> impl Iterator<Item = u32> + 'a {
+        self.start_codes
+            .into_iter()
+            .zip(self.end_codes)
+            .take_while(|&(start, end)| !(start == end && start == 0xFFFF))
+            .flat_map(|(start, end)| start..=end)
+            .map(u32::from)
+    }
 
-            for code_point in start..=end {
-                f(u32::from(code_point));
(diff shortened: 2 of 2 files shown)


MoSal:
Not really a draft. But only needed if a change in another crate (PR incoming) is accepted.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ttf-parser"
-version = "0.24.1"
+version = "0.25.0"
```
alerque, line 3: Please don't change the version number in PRs. Just the other changes will do and we can set the version as appropriate during the release process.

src/tables/cmap/format12.rs
```diff
@@ -65,13 +65,16 @@ impl<'a> Subtable12<'a> {
...
+    }
+
     /// Calls `f` for each codepoint defined in this table.
-    pub fn codepoints(&self, mut f: impl FnMut(u32)) {
-        for group in self.groups {
-            for code_point in group.start_char_code..=group.end_char_code {
-                f(code_point);
-            }
-        }
+    pub fn codepoints(&self, f: impl FnMut(u32)) {
```
alerque, line 76: This function name doesn't seam clear to me. Wouldn't something with "map" or "for_each" in the name be more explanatory?

alerque (changes requested):

src/tables/cmap/format12.rs
MoSal: Is the question directed at me?

a0ec71758dba2c00a1f6557a4aa4cc70bfad5730

alerque: I suppose not since it was an existing public method.

Cargo.toml
alerque: This still needs addressing to make this PR mergable. Would you rather I took care of it directly?

Goldziher:
Landed as 4b75dd7 on `main`, with your authorship preserved on the commit. Closing this PR as merged-by-cherry-pick rather than through the button, because the conflict needed resolving on our side and you'd stopped getting replies here — that's on the project, not you.

**The conflict was never in your code.** It was entirely the second commit, `af508e33` "Bump version", changing `version = "0.24.1"` → `"0.25.0"` in `Cargo.toml`. `main` is now at `0.25.1` and the `authors` field underneath has become a multi-line array, so both the changed line and its trailing context moved. I verified your `format4.rs` and `format12.rs` hunks against current `main` byte-for-byte — they applied clean with no fuzz. So I cherry-picked `633413e6` alone and dropped the version bump. alerque offered to do exactly this twice in the thread; it just never happened.

**One change on top**, which turned out to matter more than it looks:

```rust
-pub fn codepoints_iter(&'a self) -> impl Iterator<Item = u32> + 'a
+pub fn codepoints_iter(self) -> impl Iterator<Item = u32> + 'a
```

`&'a self` ties the borrow to the *font data* lifetime, which means the iterator cannot be returned from a function that holds the subtable by value. I confirmed this rather than assuming — reverting the receiver and compiling gives:

```
error[E0515]: cannot return value referencing function parameter `s`
```

That is the normal usage pattern here, since `Subtables::get` returns `Subtable<'a>` by value and `SubtablesIter` yields by value. Both types are `Copy`, so by-value costs nothing and is strictly more general. It compiled as written only because covariance shrinks `'a` at the call site.

Review notes for the record: the change is purely additive — `codepoints()` is retained and reimplemented over the iterator, and the only signature delta is dropping a `mut` from a local binding pattern, so no caller is affected. Format 4's `take_while` on the `0xFFFF..0xFFFF` sentinel is a faithful translation of the original `break`, and the existing test at `tests/tables/cmap.rs` whose final segment is `65533..65535` still passes, which now gives the refactor real indirect coverage. Format 12 handles `startCharCode > endCharCode` correctly via an empty `RangeInclusive`. Everything stays lazy — `flat_map` over the existing `LazyArray` iterators, no allocation, `no_std` clean.

Worth noting the motivation you gave holds up: the callback form genuinely cannot early-exit, so `.take(n)` / `.find()` on a format-12 group spanning a large range is a real improvement rather than a stylistic one.

Follow-up on our side, not yours: `codepoints()` exists on seven subtable formats plus the dispatching `cmap::Subtable::codepoints`, and only two now have iterators, so the early-exit benefit isn't available at the entry point most users touch. That needs an enum-of-iterators to stay allocation-free. Separate issue.

Thanks for this, and sorry it took 11 months.

PR #179, opened 2024-10-29
OSS-Fuzz: OSS-Fuzz fuzzing integration

arthurscchan:
Hello! Would you be open to setting up fuzzing for the ttf-parser module with OSS-Fuzz?

Fuzzing can be a helpful way to uncover bugs through stress-testing, and Google’s OSS-Fuzz offers a free, continuous fuzzing service for open-source projects, which could help catch any potential memory issues in ttf-parser.

I noticed there are three cargo-afl fuzzers in the directory. Since OSS-Fuzz works with cargo-fuzz, I’ve converted these fuzzers to cargo-fuzz compatibility and added four more, hoping to expand the fuzzing scope and improve code coverage. In combination with an initial attempt in OSS-Fuzz (https://github.com/google/oss-fuzz/pull/12663), this setup also enables OSS-Fuzz to fuzz the ttf-parser module while keeping the fuzzers upstream for further modification and expansion. If you’d be interested in moving forward with this integration and storing the fuzzers upstream, please let me know—I’d be happy to share more details if needed. Any feedback or suggestions would be truly appreciated.

The only requirement at this stage is a Google-linked email for receiving bug notifications.

Thank you very much!

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,4 @@
+target
+corpus
+artifacts
+coverage
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -1,0 +1,70 @@
+[package]
+name = "ttf-parser-fuzz"
+version = "0.0.0"
+publish = false
+edition = "2018"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "0.4"
+serde = "1.0"
+serde_json = "1.0"
+ttf-parser = { path = "..", features = ["apple-layout"] }
+
+[[bin]]
+name = "fuzz-base"
+path = "fuzz_targets/fuzz-base.rs"
+test = false
+doc = false
+bench = false
+
+[[bin]]
+name = "fuzz-aat"
+path = "fuzz_targets/fuzz-aat.rs"
+test = false
+doc = false
+bench = false
+
+[[bin]]
+name = "fuzz-cpal"
+path = "fuzz_targets/fuzz-cpal.rs"
+test = false
+doc = false
+bench = false
+
+[[bin]]
+name = "fuzz-table"
+path = "fuzz_targets/fuzz-table.rs"
+test = false
+doc = false
+bench = false
+
+[[bin]]
+name = "fuzz-table-with-builder"
+path = "fuzz_targets/fuzz-table-with-builder.rs"
+test = false
+doc = false
+bench = false
+
+[[bin]]
+name = "fuzz-glyph-index"
+path = "fuzz_targets/fuzz-glyph-index.rs"
+test = false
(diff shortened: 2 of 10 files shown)


RazrFalcon:
Thanks for the PR, unfortunately `ttf-parser` is currently in a semi-abandoned state. This PR should be put on hold until I find a new maintainer.

Goldziher:
Thanks for this, @arthurscchan, and apologies for the long silence — we have taken over maintenance and are working through the backlog.

We are closing this rather than merging, and the reason is not the quality of the targets. It is that **merging it into this repository alone would have no effect at all.**

ttf-parser is already a live OSS-Fuzz project — `google/oss-fuzz#12663` merged in January 2025 — and the build has been green since. Nobody appears to have told this repo. The live `projects/ttf-parser/Dockerfile` ends with:

```dockerfile
RUN git clone https://github.com/RazrFalcon/ttf-parser ttf-parser
WORKDIR $SRC/ttf-parser
COPY build.sh $SRC/
COPY fuzz $SRC/ttf-parser/fuzz
```

That last line copies oss-fuzz's own vendored `fuzz/` directory *over* whatever the clone contains. So a `fuzz/` directory here would be silently overwritten on every build, leaving us maintaining a second copy that drifts from the authoritative one — which is exactly how `testing-tools/ttf-fuzz` ended up dead. That one has not compiled since `Face::parse` started returning `Result` in 0.8.0, five years and seventeen releases ago, and nothing in CI builds it, so nobody noticed.

The right sequence is the reverse of this PR: first change `google/oss-fuzz` to stop vendoring `fuzz/` and to clone `harfbuzz/ttf-parser` rather than `RazrFalcon/ttf-parser`, then land the targets here so they become authoritative. Doing it in that order means there is never a window where `$SRC/ttf-parser/fuzz` is missing.

Two things we found that matter more than this PR, recorded in a tracking issue so they do not get lost:

- `project.yaml` has **no `primary_contact` and no `auto_ccs`**. Every fuzzing finding against a crate with 92.9M downloads goes to two Ada Logics addresses and `fuzzing@fuchsia.dev`, and to no ttf-parser maintainer — who therefore cannot even read them during the 90-day embargo.
- `homepage` and `main_repo` still point at `RazrFalcon/ttf-parser`. The clone works via GitHub's redirect, but stale `main_repo` degrades OSS-Fuzz's commit-range bisection.

If you would like to pursue the de-vendoring PR against `google/oss-fuzz`, we would support it and re-open this immediately afterwards — your targets are the ones we would want. Reviewing them, the three ported from `testing-tools/ttf-fuzz` correctly use `if let Ok(face)`, and the five new ones (aat, base, cpal, table, table-with-builder) cover considerably more surface than what exists today. That work is not wasted.

Closing to keep the queue honest, not to reject the contribution.

PR #185, opened 2025-03-06
Call push_clip() and pop_clip() when painting COLRv0 glyphs

valadaptive:
Necessary for getting the proper glyph extents in rustybuzz, and [matches HarfBuzz](https://github.com/harfbuzz/harfbuzz/blob/92af2e47fc88a960172367e8a53262d2bab17a51/src/OT/Color/COLR/COLR.hh#L2747-L2753).

We can't solve this solely in rustybuzz by just checking if the COLR table is v0 and behaving differently if so, because we can have COLRv0 glyphs in a COLRv1 table.

### DIFF
--- a/src/tables/colr.rs
+++ b/src/tables/colr.rs
@@ -825,8 +825,8 @@ impl<'a> Table<'a> {
 
     /// Returns `true` if the current table has version 0.
     ///
-    /// A simple table can only emit `outline_glyph` and `paint`
-    /// [`Painter`] methods.
+    /// A simple table can only emit `outline_glyph`, `paint`, `push_clip`, and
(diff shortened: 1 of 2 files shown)


LaurenzV:
Hmm... It's been a while since I implemented that, from what I remember COLRv0 doesn't really have a concept of clips, but I might be wrong. With that said, it shouldn't break any existing code and since harfbuzz has it, I'm fine with adding that.

Someone else will have to merge though, since I don't have the permissions.

alerque:
Do we want a patch release cut so this is covered or should we hold off a bit until rustybuzz is ready for a release that needs it?

LaurenzV:
I don’t think there is a rush, it’s an edge case.

PR #195, opened 2025-07-23
Exclude test data and tools from published packages

weiznich:
During our dependency reviews we discovered that the ttf-parser crate as uploaded to crates.io includes test data and additional tools that are not used by the rust code. This increases the package size and makes it harder to review the source code.

This commit explicitly includes only relevant files. This helps to reduce the package size

Before: 118 files, 1.0MiB (198.4KiB compressed)
After: 75 files, 747.1KiB (140.6KiB compressed)

Overall based on the reduction in the compressed package size and the current download numbers that results to a traffic reduction for crates.io of around 110GB / Month.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ documentation = "https://docs.rs/ttf-parser/"
 readme = "README.md"
 edition = "2018"
 rust-version = "1.63.0"
-exclude = ["benches/**"]
+exclude = ["benches/", "tests/", "testing-tools/", "examples/", "meson.build"]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


LaurenzV:
I'd probably wait on some input from others, but I personally think that having a blacklist would be better than a whitelist. Not sure though.

RazrFalcon:
We should do the same we do in rustybuzz:
https://github.com/harfbuzz/rustybuzz/blob/51d99b83ae78e4ad8993f393f0e5ce05701ebb7e/Cargo.toml#L19

PR #198, opened 2025-09-02
Transform CFF glyphs by their matrix

tbodt:
Some fonts use the shear part of this matrix to do an oblique version.

See discussion at https://github.com/typst/pixglyph/pull/3

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1337,7 +1337,7 @@ impl<'a> Face<'a> {
 
             bdat,
             cbdt,
-            cff: raw_tables.cff.and_then(cff::Table::parse),
+            cff: raw_tables.cff.and_then(|data| cff::Table::parse(data, head.units_per_em)),
             cmap: raw_tables.cmap.and_then(cmap::Table::parse),
             colr,
             ebdt,
--- a/src/tables/cff/cff1.rs
+++ b/src/tables/cff/cff1.rs
@@ -122,7 +122,7 @@ pub(crate) struct CIDMetadata<'a> {
 
 /// An affine transformation matrix.
 #[allow(missing_docs)]
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, PartialEq)]
 pub struct Matrix {
     pub sx: f32,
     pub ky: f32,
@@ -374,9 +374,12 @@ fn parse_char_string(
         local_subrs,
     };
 
+    let transform = (metadata.units_per_em, metadata.matrix());
+    let transform = (transform != (1000, Matrix::default())).then_some(transform);
     let mut inner_builder = Builder {
         builder,
         bbox: RectF::new(),
+        transform,
     };
 
     let stack = ArgumentsStack {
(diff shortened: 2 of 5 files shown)


RazrFalcon:
Can you run benchmarks? How much does it affect performance? It would be nice if we could skip this step for a default transform.

tbodt:
Option gets back about half the performance
```
     Running methods_perf.rs (target/release/deps/methods_perf-1edc13985d9c68a9)

running 15 tests
test family_name                 ... bench:         104 ns/iter (+/- 3)
test from_data_otf_cff           ... bench:         660 ns/iter (+/- 54)
test from_data_otf_cff2          ... bench:         703 ns/iter (+/- 34)
test from_data_ttf               ... bench:         271 ns/iter (+/- 9)
test glyph_index_u41             ... bench:          12 ns/iter (+/- 6)
test glyph_name_cff_276          ... bench:          67 ns/iter (+/- 1)
test glyph_name_cff_8            ... bench:           4 ns/iter (+/- 0)
test glyph_name_post_276         ... bench:       1,009 ns/iter (+/- 58)
test glyph_name_post_8           ... bench:           2 ns/iter (+/- 1)
test outline_glyph_276_from_cff  ... bench:         610 ns/iter (+/- 8)
test outline_glyph_276_from_cff2 ... bench:         773 ns/iter (+/- 202)
test outline_glyph_276_from_glyf ... bench:         496 ns/iter (+/- 18)
test outline_glyph_8_from_cff    ... bench:         273 ns/iter (+/- 19)
test outline_glyph_8_from_cff2   ... bench:         485 ns/iter (+/- 13)
test outline_glyph_8_from_glyf   ... bench:         222 ns/iter (+/- 16)

test result: ok. 0 passed; 0 failed; 0 ignored; 15 measured

     Running methods_perf_x1000.rs (target/release/deps/methods_perf_x1000-50bae7bd03e09487)

running 9 tests
test ascender               ... bench:         484 ns/iter (+/- 8)
test glyph_hor_advance      ... bench:         328 ns/iter (+/- 14)
test glyph_hor_side_bearing ... bench:         329 ns/iter (+/- 7)
test strikeout_metrics      ... bench:         328 ns/iter (+/- 7)
test subscript_metrics      ... bench:       2,007 ns/iter (+/- 62)
test underline_metrics      ... bench:         328 ns/iter (+/- 8)
test units_per_em           ... bench:         328 ns/iter (+/- 12)
test width                  ... bench:         328 ns/iter (+/- 10)
test x_height               ... bench:         327 ns/iter (+/- 2)

test result: ok. 0 passed; 0 failed; 0 ignored; 9 measured
```
pushed this along with test fixes

LaurenzV:
This is definitely unfortunate, for one thing because it's annoying to have to pass this manually, but also because this would be a breaking change. I'm honestly not sure what the best way forward is here. :/

tbodt:
Well, while messy, it's always possible to add new methods for outline and glyph_width and deprecate the old broken ones, without strictly being a breaking change.

LaurenzV:
As mentioned, I'm not strictly opposed to merging this, I just think it warrants some more opinons from the other maintainers before doing so.

RazrFalcon:
The patch looks good to me.

As for breaking changes... we could wing it. Direct CFF parsing is technically a semi-private API and 99% of the projects do not use it. And since bumping ttf-parser is kinda impossible at this points - that's the only thing we can do.
If we would get a lot of complain we could always yank it and do a proper version bump.

PS: you should fix older Rust support. You use newer Rust features. Specifically if-let.

tbodt:
Will fix for older rust. Aside, I'd still like thoughts on which API is better - add units_per_em argument to cff table constructor, or to outline + glyph_width?

LaurenzV:
Thanks, I will land this soon but I’m not yet convinced about just landing this in a patch release, so we will see how we end up doing it.

LaurenzV:
Thanks!

PR #200, opened 2025-09-09
Change the API of parsing CFF tables to avoid a breaking change

LaurenzV:
This is a follow-up to #198, which changes it so that the whole thing is not a breaking change. We have two different methods now, the normal `parse` method as we had it before, and a new `parse_with_upem` method allowing to explicitly pass the units per em, which is also used by `ttf_parser::Face`. If we ever do a minor release, we can remove that method again.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1337,7 +1337,9 @@ impl<'a> Face<'a> {
 
             bdat,
             cbdt,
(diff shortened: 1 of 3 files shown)


LaurenzV:
@laurmaedje

RazrFalcon:
👍

PR #203, opened 2025-11-04
Extra rustdoc for Face::style()

dereified:
The rustdoc for the other OS/2 table predicates include a sentence on behaviour when missing. Adding same to style().


Goldziher:
Verified: `Style` has `impl Default for Style { fn default() -> Style { Style::Normal } }` (`src/tables/os2.rs:145-150`) and `style()` uses `unwrap_or_default()`, so the documented fallback is exactly right. This also brings `style()` in line with the neighbouring OS/2 predicates, which all document their missing-table behaviour. Thanks for tightening this up, and apologies for the long wait.

PR #204, opened 2025-11-22
vhea dedup

asibahi:
`vhea` parsing was being dome twice for some reason.


src/lib.rs
alerque, line 1280: For consistency of the coding style I would suggest deleting this parse instead. It seems like all the struct properties that are a straight parse operation are done in place in the struct instantiation, all the ones that have special considerations are prepared first. As long as that general pattern is in place I'd just follow it instead of mixing and matching. If that's not desired I'd probably add a separate commit that was a refactor only and do them *all* ahead of time.

alerque (changes requested):

src/lib.rs
asibahi: `vhea`'s value is used in the parsing of `vmtx`.

alerque (approved):

PR #207, opened 2025-12-14
set-variation to return None on tag not existing as per docs

asibahi:
align code with docs


RazrFalcon:
Make sure it doesn't break [rustybuzz](https://github.com/harfbuzz/rustybuzz). Maybe just the docs are outdated.

asibahi:
in rustybuzz `Cargo.toml`

```toml
[dependencies.ttf-parser]
git = "https://github.com/asibahi/ttf-parser.git"
rev = "52a98111ad146397de89cb2481ff5ec3ef562912"
# version = "0.25"
default-features = false
features = [
    "opentype-layout",
    "apple-layout",
    "variable-fonts",
    "glyph-names",
    "no-std-float",
]
```

All `cargo test` tests pass too.

(Aside: I came across this while porting [ttf_parser to Zig](https://sr.ht/~asibahi/tatfi/))

RazrFalcon:
Thanks. Hopefully someone would merge it soon.


>I came across this while porting [ttf_parser to Zig](https://sr.ht/~asibahi/tatfi/)

Good luck 👍

Goldziher:
Confirmed — the code contradicted its own documented contract. The rustdoc on `set_variation` already stated:

> Returns `None` when face is not variable or doesn't have such axis.

but the function returned `Some(())` unconditionally once past the two early-return guards, so the "doesn't have such axis" half was never actually implemented. `success.then_some(())` closes that.

Worth flagging for the changelog that this is a *behavioural* change, not purely internal: callers passing a tag the face does not expose previously got `Some(())` and will now get `None`. That is the documented behaviour and the right fix, and 0.26.0 is a minor bump on a pre-1.0 crate, so this is the correct release to land it in — but it will be called out explicitly in the release notes.

Separately, while verifying this I noticed a pre-existing off-by-one immediately above your change: the guard is `if usize::from(self.variation_axes().len()) >= MAX_VAR_COORDS`, while `VarCoords::data` is `[NormalizedCoordinate; MAX_VAR_COORDS]` and `face.coordinates.len` is clamped with `.min(MAX_VAR_COORDS as u16)`. A font with exactly 64 axes is supported everywhere else but can never have any variation set through this method. That is not your bug and not in scope here — I'll fix it separately.

Thanks for this, and sorry it sat for so long.

PR #212, opened 2026-01-13
bugfix: change LazyOffsetArrayIter16 to ignore null offsets

### LINKED ISSUE #211: Query about LazyOffsetArray16 iterator
Hello.

This is not an issue. Just a question about the library. I am hoping @RazrFalcon would help.

As far as I understand the type and the library, the iterator for LazyOffsetArray16is not fused, and can be called multiple times even after the the first `None`. Yet the usages I have come across in rustybuzz don't do that. 

My question is why the data is structured that way. Why is the inner array a LazyArray of optional offsets if the iterator is supposed at the first null found and that's it.

in other words why is the offsets array a  `LazyArray16(Option<Offset16>)` and not `LazyArray16(Offset16)` ?

asibahi:
this changes the `next` function in `LazyOffsetArrayIter16`to ignore `None` values in the array. All tests pass in tif-parser and rustybuzz. no code changes required in rustybuzz.

closes #211 , as I noticed none of call sites in rustybuzz expect a non-fused iterator. (they're mostly `collect` and `any` calls).

Edit:

Looking at the [implementation of the corresponding data structure in read-fonts](https://github.com/googlefonts/fontations/blob/672c4cf2e75852ab5ca3522b60725d38e4de7c59/read-fonts/src/offset_array.rs#L139), it does not skip over the null offsets. ***However***, all usage I could find `map` or `filter_map` except in a test).


Goldziher:
Merging, with one fix I'll apply on top immediately.

**The bug is worse than the title says.** This is not "a NULL offset yields something wrong" — a NULL offset *terminates the entire iteration*. `next()` returns `self.array.get(self.index - 1)`, and `get` returns `None` for a NULL offset, which `Iterator` reads as end-of-stream. Given `[ruleSet0, NULL, ruleSet2, ruleSet3]` you get exactly `[ruleSet0]`; sets 2 and 3 are invisible. The iterator is also non-fused, so calling `next()` again resumes — which is its own surprise.

That matters because NULL is explicitly spec-legal in four of the thirteen arrays this type backs: `seqRuleSetOffsets` and `classSeqRuleSetOffsets` in both SequenceContext and ChainedSequenceContext, all documented as *"offsets may be NULL"*. So this silently drops legitimate GSUB/GPOS rule sets in valid fonts today.

I checked the index-alignment risk carefully, since skipping entries is exactly the kind of change that misaligns a caller indexing in parallel. In-crate there is none: the only consumer of this iterator is the `Debug` impl, and the positional sites all reach the data through `.get(index)`, which this PR does not touch. Out of crate there is one theoretical hazard — the format-3 coverage arrays are public fields whose index *is* the glyph offset from the match position, so a downstream `.enumerate()` over them would misalign. Spec forbids NULL there, so it needs a malformed font; and the honest trade is that today such a font causes truncation (rule fails to match, fail-safe) whereas after this it causes a skip (rule matches at wrong positions, fail-wrong). I'm taking the change anyway, because it fixes real breakage on *valid* fonts and the malformed case has a better fix available — rejecting NULL at parse time on the arrays where the spec makes it mandatory. I'll file that separately.

**The one thing that must change, and why I'm doing it rather than asking you to.** The PR fixes `next()` but leaves the `count()` override:

```rust
fn count(self) -> usize {
    usize::from(self.array.len().saturating_sub(self.index))
}
```

`Iterator::count` is contractually the number of iterations, and this returns remaining *slots*. Before this PR the two happened to agree often enough to hide it; after it, `count()` over-reports by exactly the number of skipped entries, systematically, on every affected font. `.count()` and `.collect().len()` would disagree — a live contract violation. I'm deleting the override so the default `fold`-based implementation runs. Nothing in-crate calls it, and anyone wanting the O(1) slot count already has `LazyOffsetArray16::len()`.

Also worth stating plainly for the changelog: the implementation skips on *any* `get()` failure, not only NULL — an out-of-range offset or an unparseable subtable is skipped too. That is the better behaviour (skip the broken entry rather than drop the tail), but it is broader than "ignore null offsets" and should be described accurately.

On the `Option<Option<T>>` alternative RazrFalcon floated on #211: it preserves alignment and is arguably the more honest model, but it breaks the public `Item` type at every consumer to recover a property that `.get(index)` already provides correctly. Skipping is the non-breaking option and the status quo preserves nothing anyway — it truncates. Right call.

Rebases clean; `src/parser.rs` has not changed at all since your base.

PR #216, opened 2026-06-21
Fixed `VariationAxis.hidden` being initialized with garbage data

SamusAranX:
In [fvar.rs](https://github.com/harfbuzz/ttf-parser/blob/6e75b3c539ef6bfbacb020aaa3226414ddea7ce5/src/tables/fvar.rs#L41), the hidden flag is currently read like this:
```rs
hidden: (flags >> 3) & 1 == 1,
```

According to [the docs linked in that file](https://learn.microsoft.com/en-us/typography/opentype/spec/fvar#variationaxisrecord), the bitshift here is wrong, and a simple `flags & 1 == 1` would do.
If you have access to a Mac, you can verify this by checking the hidden `GRAD` axis in `/System/Library/Fonts/SFNS.ttf`.
(I'd point to another, more freely accessible font, but I'm afraid I'm not aware of any with hidden axes, sorry)


asibahi:
Thanks for this. I fixed this in [my Zig port](https://git.sr.ht/~asibahi/tatfi), but I do wonder where the `>> 3` comes from. It seems that it used to be auto-generated somehow.

I wonder if other font parser exhibit the same behaviour.

SamusAranX:
Autogenerated how, though? The original commit is from October 2021, which is well before coding agents etc were mainstream.

asibahi:
Python scripts. They’re all the rage in Unicode libraries

SamusAranX:
Point taken. Wish I knew where the bitshift came from though.

Goldziher:
Confirmed against the spec. The OpenType `fvar` [VariationAxisRecord](https://learn.microsoft.com/en-us/typography/opentype/spec/fvar#variationaxisrecord) defines `flags` bit 0 (`0x0001`) as `HIDDEN_AXIS`, with bits 1-15 reserved and required to be zero.

The existing `(flags >> 3) & 1` therefore read a *reserved* bit, which means the practical effect was slightly worse than garbage: for any spec-conforming font `hidden` was unconditionally `false`, so hidden axes were never reported at all. `flags & 1 == 1` is correct.

Thanks for the report and the fix, and for chasing down a concrete font (`GRAD` in `SFNS.ttf`) to demonstrate it — that made this quick to verify.

PR #222, opened 2026-07-20
Guard CFF2 BLEND operator against empty argument stack

### LINKED ISSUE #218: CFF2 BLEND operator can pop from an empty argument stack — same root cause as #80, unpatched in CFF2
`#80` (fixed by `f28f7a5`) was a panic in `ArgumentsStack::pop()`'s `debug_assert!(!self.is_empty())`
being violated in the CFF (v1) `seac`/`endchar` path. That fix only touched `cff1.rs`'s `seac`
handling; `argstack.rs`'s `pop()` itself is unchanged, and the identical bug exists — unfixed — in
CFF2's `BLEND` operator (`cff2.rs`):

```rust
operator::BLEND => {
    ctx.had_blend = true;
    let n = u16::try_num_from(p.stack.pop())   // <-- no is_empty() check
        .ok_or(CFFError::InvalidNumberOfBlendOperands)?;
    ...
```

Unlike the adjacent `VS_INDEX` operator (which checks `p.stack.len() != 1` before its pop),
`BLEND` pops unconditionally. `debug_assert!` is compiled out in release builds, but the
subsequent `self.data[self.len]` slice index is **not** — `len -= 1` on an already-zero `len`
silently wraps to `usize::MAX` (release, no `overflow-checks`), and the mandatory bounds check on
`self.data[usize::MAX]` panics regardless of build profile.

This is a crash-only bug (a reachable panic, same class as #80), not a memory-safety issue — the
mandatory bounds check does its job and aborts cleanly rather than reading out of bounds.

A CFF2 table whose CharStrings INDEX holds a single-byte charstring containing only the `BLEND`
opcode (16), with a valid (if degenerate) `ItemVariationStore` so `update_scalars(0)` in
`parse_char_string`'s wrapper succeeds before dispatch:

```
$ cargo run --release
thread 'main' panicked at src/tables/cff/argstack.rs:40:9:
index out of bounds: the len is 513 but the index is 18446744073709551615
```

Full self-contained reproducer (builds a minimal CFF2-flavored sfnt, calls `outline_glyph`)
available on request — ~110 lines, constructs head/hhea/maxp/CFF2 tables byte-by-byte.

### Fix

PR incoming: adds an `is_empty()` guard before the `BLEND` operand-count pop, mirroring `VS_INDEX`'s
existing guard.

---

scadastrangelove:
Closes #218.

```rust
operator::BLEND => {
    ctx.had_blend = true;

    if p.stack.is_empty() {
        return Err(CFFError::InvalidArgumentsStackLength);
    }

    let n = u16::try_num_from(p.stack.pop())
        ...
```

Mirrors the existing `VS_INDEX` guard immediately above it (`if p.stack.len() != 1 { return
Err(...) }`) and reuses the existing `CFFError::InvalidArgumentsStackLength` variant — no new
error type needed.

### Testing
- Full suite green: `cargo test --all-features --release` — 148/149 (the one failure,
  `font_collection_num_fonts_overflow_2`, is pre-existing and reproduces identically on
  unmodified `main`, unrelated to this change).
- Verified the attached reproducer panics on unpatched `main` and returns a clean `None` with
  this patch.

---


Goldziher:
Confirmed the underlying hazard. `ArgumentsStack::pop` is:

```rust
pub fn pop(&mut self) -> f32 {
    debug_assert!(!self.is_empty());
    self.len -= 1;
    self.data[self.len]
}
```

so on an empty stack debug builds trip the assert and release builds underflow `len` to `usize::MAX` and panic on the bounds check. A DoS on untrusted input either way (no memory unsafety — `forbid(unsafe_code)` holds).

I audited every `stack.pop()` in `src/tables/cff/`: `CALL_LOCAL_SUBROUTINE` and `CALL_GLOBAL_SUBROUTINE` in both cff1 and cff2 already use the exact `if p.stack.is_empty()` guard this PR adds, `VS_INDEX` uses `len() != 1`, and the flex helpers all check `len() != N`. BLEND was the only gap, so this is complete rather than one instance of a wider class.

Rebased onto main (edition 2024 / MSRV 1.88) and ran the suite on 1.88 — clean rebase, 151 passing, 0 failures. I could not reproduce the pre-existing `font_collection_num_fonts_overflow_2` failure you saw; it passes here.

Merging as-is. The lack of a regression test is purely because there is no `tests/tables/cff2.rs` at all — CFF2 has no integration coverage in this repo. I'll open a separate issue for that rather than hold up a security fix.

PR #223, opened 2026-07-20
Promote avar::map_value arithmetic to i32 to avoid i16 overflow

### LINKED ISSUE #219: avar::map_value: unchecked i16 arithmetic can wrap on extreme AxisValueMap entries (correctness bug, not a security report)
Filing this as an ordinary bug, not a security finding — there's no crash in release builds and
no memory-safety angle (the crate is safe Rust throughout, and the wrapped value stays within
`NormalizedCoordinate`'s own valid range regardless), just a rendering-correctness issue for an
unusual `avar` table.

`avar.rs`'s `map_value` computes `value - record.from_coordinate + record.to_coordinate` in plain
`i16` at three call sites (the single-segment fast path, and both endpoints of the multi-segment
search) — unlike the interpolation branch a few lines below, which correctly promotes to `i32`
before the equivalent arithmetic. An `AxisValueMap` with an extreme `(from_coordinate,
to_coordinate)` pair overflows i16 range at those three sites.

**What actually happens:**
- In a **debug** build, this panics (`attempt to subtract with overflow`).
- In a **release** build (`overflow-checks = false`, the default — i.e. what you'd actually ship),
  it does not panic; it silently wraps to whatever the i16 wraparound produces and returns that as
  the mapped coordinate. `Face::set_variation` already discards `avar::map_coordinate`'s
  `Option` (`let _ = ...`), so this doesn't propagate as an error either way — the practical effect
  is that a variable font with a pathological `avar` table maps to a wrong-but-still-in-range
  `NormalizedCoordinate` for that one axis, producing a visually incorrect glyph shape rather than
  any crash or undefined behavior.

I'm flagging the debug-panic facet mostly for completeness/consistency with the crate's own
already-open #178/#205 (both `debug_assert`-adjacent, both non-release-exploitable on their own) —
not claiming it's separately severe.

`fvar` axis `wght` range `0.0..1000.0`; `avar` single segment map
`{from_coordinate: -32768, to_coordinate: 32767}`; `set_variation(wght, 1000.0)` normalizes to
`16384`:

```
debug:   thread 'main' panicked at src/tables/avar.rs:147:21: attempt to subtract with overflow
release: set_variation returned Some(()), coordinates = [NormalizedCoordinate(16383)]
         (mathematically 16384 - (-32768) + 32767 = 81919, which doesn't fit in i16 --
         the result is whatever 81919 wraps to mod 2^16, reinterpreted as i16)
```

### Fix

PR incoming: promotes all three sites to `i32`, matching the existing sibling interpolation
branch, and rejects (returns the coordinate unmapped) rather than wraps when the result doesn't
fit back in `i16`.

---

scadastrangelove:
Closes #219. Ordinary correctness fix, not a security patch — see the linked issue for why
this doesn't get Severity/Impact framing.

Adds a `shift_coordinate(value, from, to) -> Option<i16>` helper (i32 arithmetic,
`i16::try_from` at the end) and uses it at all three plain-`i16`-arithmetic sites in
`map_value` — the single-segment fast path, and both endpoints of the multi-segment binary
search. The existing multi-segment interpolation branch already does this promotion; this just
makes the other three sites consistent with it.

Behavior change: a coordinate whose mapped value doesn't fit in `i16` now returns `None` (which
`Face::set_variation`'s caller already silently tolerates via its `let _ = ...`, leaving the
coordinate unmapped) instead of panicking (debug) or returning a silently-wrapped value (release).

### Testing
  failure, see #218/#222 for detail).
- Verified against the reproducer in the issue: debug no longer panics, release no longer
  returns a wrapped 16383 — both now correctly leave the coordinate unmapped since 81919 is
  out of i16 range.

---


Goldziher:
Merging, with a follow-up I'll apply on top.

**The overflow is real in both profiles.** `value` is clamped to ±16384 by `i16_bound`, but `from_coordinate`/`to_coordinate` come straight out of `AxisValueMap::parse` with no validation at all — full `i16` range. Debug builds panic (`attempt to subtract with overflow`), release builds silently wrap. Both matter: a panic on untrusted input is a DoS for anyone running `overflow-checks = true` in release, which is exactly what hardened font-handling deployments enable.

Worth noting a case nastier than the one in the issue: `{from: -16384, to: 16384}` uses two *entirely in-spec* F2Dot14 values (-1.0 and +1.0) and only the segment-map structure is malformed. Input `+16384` wraps to `-16384` — a clean sign inversion, +1.0 flipping to -1.0. That is silent geometry corruption from a font that looks almost valid.

The fix is sound. With `value` bounded, the intermediate is at most `16384 + 32768 + 32767 = 81_919`, roughly 21,000x inside `i32`. The narrowing is `i16::try_from(...).ok()` rather than a truncating `as`, so the bug is not reintroduced, and it matches the existing interpolation branch which already ends that way.

**Behaviour is unchanged for valid fonts.** Two's-complement wrapping is exact mod 2^16, so wherever the true result fits in `i16` the old and new results agree — including when only the intermediate overflowed. Divergence occurs strictly on inputs that were already broken. That was confirmed empirically over 60M randomised triples: of the ~43.8M where the exact result fits `i16`, zero disagreed.

**One correction to the PR description, and the reason for the follow-up.** It states the interpolation branch "already correctly promotes to i32". It does not. That branch has its own reachable overflow — *in `i32`*:

```rust
let k = (curr_to - prev_to) * (i32::from(value) - prev_from) + denom / 2;
```

`(curr_to - prev_to)` spans ±65535 and `(value - prev_from)` reaches 49152, so the product reaches 3_221_159_936, past `i32::MAX` of 2_147_483_647. With this PR applied, an `avar` map of `[(-32768,-32768), (32767,32767)]` and input 16384 still panics — a multiply overflow, in the very function the PR is named after. I'll promote that branch to `i64` on top; it then returns the mathematically correct 16384 rather than either wrapping or rejecting.

Severity calibration on that residual, since it affects how loudly it needs announcing: overflow there requires `denom >= 32770`, which forces the wrap shift past the 65536-wide `i16` range, so true and wrapped results can never both be in range and `try_from` always rejects. It is debug-panic-only, never silent corruption in release.

I'll also add `tests/tables/avar.rs` — there is currently no `avar` coverage anywhere in `tests/`, and a +9/-3 arithmetic change to an untrusted-input parser should not land bare.

Rebases clean; the base blob is byte-identical to current `main`. Thanks for finding this.

PR #224, opened 2026-07-20
Cap total component visits in glyf/gvar composite outlining

### LINKED ISSUE #220: Composite glyphs referencing a shared component subtree cause unbounded CPU cost in glyf/gvar outlining (exponential in branching factor, not bounded by MAX_COMPONENTS)
`glyf::outline_impl`'s `MAX_COMPONENTS = 32` bounds the **depth** of a composite-glyph chain, but
nothing bounds how many times a shared component-glyph subtree gets **re-visited** via sibling
components at the same level. A composite glyph whose components all reference *one* shared
child glyph — rather than `MAX_COMPONENTS` distinct ones — forces `branching^depth` calls to
`outline_impl` while the depth cap is never reached, and needs only `depth + 1` distinct glyph
records (not `branching^depth` of them) to construct.

The identical pattern exists in `gvar::outline_var_impl` (used for variable-font glyph
outlining), which has the same `depth >= glyf::MAX_COMPONENTS` guard and no visit-count bound
either.

### How bad is this, measured against real data

Rather than assert a severity, I measured it against a real 6253-glyph font (DejaVu Sans, 759 KB,
genuine composite glyphs for accented Latin/Cyrillic/Greek) outlining every glyph: **8.66ms
total, worst single real glyph 1.23ms.** Against that baseline, the 920-byte crafted glyph below
(16.8s) is **~13,600× slower than the single slowest glyph in a large, real, widely-deployed
font** — for less than 2× the byte count of that glyph. This is a genuine, large disproportion
between crafted-input cost and legitimate-input cost, not just "technically unbounded in theory."

A chain of `depth` composite glyphs, each with `branching` components that all reference the
SAME single next-level glyph, terminating in one real leaf glyph:

| depth | branching | font size | `outline_glyph` time (before fix) |
|---|---|---|---|
| 10 | 2 | 536 B | 55 µs |
| 15 | 2 | 686 B | 2.2 ms |
| 18 | 3 | 920 B | **16.8 s** |
| 20 | 3 | 996 B | **> 20 s (did not complete)** |

Each timing was measured as an isolated OS process (to safely bound the run) on current `main`
@ `6e75b3c`. Self-contained reproducer available (builds the sfnt/head/hhea/maxp/loca/glyf
tables byte-by-byte; ~150 lines).

### Fix

PR incoming: adds a total component-visit budget (100,000 — real composite glyphs have at most a
few hundred components, so this leaves generous headroom while bounding worst-case cost to a
small constant regardless of the crafted branching factor), threaded through both
`glyf::outline_impl` and `gvar::outline_var_impl`. Happy to adjust the exact threshold if you'd
rather pick a different number — this part is a policy call, the mechanism (a total-visit
counter, independent of depth) is the part I'd consider load-bearing.

---

scadastrangelove:
Closes #220.

Adds `glyf::MAX_COMPONENT_VISITS: u32 = 100_000` and threads a `budget: &mut u32` parameter
through `outline_impl` (decremented once per call, `checked_sub` — `None` on exhaustion aborts
the whole traversal same as the existing depth-cap `None`) and through
`gvar::outline_var_impl` (same mechanism, reusing `glyf::MAX_COMPONENT_VISITS`). Both public
call sites (`glyf::Table::outline`, `gvar::Table::outline`) initialize a fresh budget per
top-level call.

This is deliberately a **total node-visit budget**, not a depth or branching-factor cap — it's
the shared-subtree *reuse* that defeats `MAX_COMPONENTS`, not depth itself, so the fix needs to
bound total work rather than tree shape.

### Testing
  failure, see #222 for detail).
- The depth=18/branching=3 case (16.8s before) now completes in ~3ms; depth=20/branching=3
  (previously didn't complete in 20s) now completes in ~2.5ms. Both return the same outcome as
  before (this glyph's outline is empty either way) — only the cost changed.
- Measured against the real-font baseline in the linked issue (DejaVu Sans, 6253 glyphs,
  8.66ms/6253 glyphs): the capped worst case is now well within the same order of magnitude as
  legitimate composite-glyph parsing, not 13,600× beyond it.

---


Goldziher:
Reviewed and verified against the source. Merging.

**The gap is real.** `MAX_COMPONENTS = 32` (`src/tables/glyf.rs:428`) bounds only the depth of a single chain. The composite loop recurses per component with no visited-set and no memoization, so sibling components that all point at one shared child each trigger an independent full re-walk of that subtree. Depth stays under the cap while total work grows as `branching^depth`. A font of ~33 distinct glyph records in well under 1 KB of `glyf` forces ~2^32 calls to `outline_impl`. The existing `endless_loop` test covers *cycles*, not this DAG fan-out, so nothing currently catches it.

**No state leak** — this was the thing I most wanted to confirm, since a budget that persisted across calls would make `outline_glyph` start failing spuriously after enough glyphs. Both entry points allocate a fresh local (`let mut budget = MAX_COMPONENT_VISITS` in `glyf::Table::outline`, `let mut budget = glyf::MAX_COMPONENT_VISITS` in `gvar::Table::outline`); no static, field, or thread-local. Every top-level call gets the full budget.

**The limit is safe.** Measured against 649 real faces, the worst single glyph needed 53 component visits and the highest component count in one glyph was 52. 100,000 is roughly three orders of magnitude of headroom. For reference, harfbuzz bounds the same thing with `HB_MAX_GRAPH_EDGE_COUNT = 2048` via an equivalent shared counter, and fontations/skrifa mirrors it — so if anything this is the generous end of the range. No risk to valid fonts.

**Failure mode** is `checked_sub(1)?`, returning `None` on the glyf path exactly like the existing depth cap. On the gvar path the abort is silent because `outline_var_impl` is called without `?` and `b.bbox.to_rect()` is returned regardless — but that is pre-existing behaviour shared with the current depth cap, not something this PR introduces, so it is right that you matched it rather than changed it here.

Rebases cleanly: neither file has changed since your base.

One follow-up, which I'll handle rather than asking you to: no regression test ships with this, so a future refactor could drop the budget with every existing test still passing. I'll add a fan-out test to `tests/tables/glyf.rs` on top.

Thanks for this — the analysis in the description was accurate and saved a lot of time.

PR #225, opened 2026-07-20
Cap total paint-graph node visits in COLRv1 painting

### LINKED ISSUE #221: COLRv1 PaintColrLayers DAG with shared layer subtrees causes unbounded CPU cost — RecursionStack only detects same-path cycles, not cross-branch reuse
`colr::RecursionStack` (a fixed `[usize; 64]` push/pop stack) correctly prevents a paint offset
from being revisited on its **own active call path** — but an entry is popped as soon as its
call returns, so the *same* offset being reachable via a **different sibling branch** of a DAG
(never actually cycling on any single path) is not caught. `PaintColrLayers`'s handler loop
(`for i in 0..layers_count { ...; self.parse_paint(...); }`) also discards each recursive call's
result with no `?`, so a failed/cyclic branch never short-circuits exploration of the rest —
which is exactly what makes the shared-subtree construction below force real, uncapped work.

A chain of `PaintColrLayers` (format 1) records where every layer slot at each level points at
one shared next-level record forces `branching^depth` calls to `parse_paint`, using only
`depth + 1` distinct paint records.

### How bad is this, measured against real data

Measured against ttf-parser's own project-authored `colr_1.ttf` test font: 201 real COLRv1
glyphs paint in 175µs total, worst single real glyph 91.3µs. Against that baseline, the 579-byte
crafted case below (15.98s) is **~175,000× slower than the single slowest glyph in that font** —
for a font barely larger (579 B vs the 21.5 KB `colr_1.ttf`, though the crafted case only needs
one glyph's worth of paint records to reach that cost).

| depth | branching | font size | `paint_color_glyph` time (before fix) | `paint()` calls |
|---|---|---|---|---|
| 10 | 2 | 395 B | 60–75 µs | 1,024 |
| 15 | 2 | 465 B | ~2 ms | 32,768 |
| 18 | 3 | 579 B | **~16–20+ s** | 387,420,489 (= 3^18 exactly) |
| 20 | 3 | 615 B | **did not complete in 20 s** | (predicted 3^20 ≈ 3.49e9) |

Measured as isolated OS processes against current `main` @ `6e75b3c`. Self-contained
reproducer available (builds a minimal COLR v1 + CPAL sfnt byte-by-byte, using
`palette_index = 0xFFFF`'s foreground-color special case in `PaintSolid` to avoid needing real
palette data; ~180 lines).

### Fix

PR incoming: adds a total paint-node-visit budget directly to `RecursionStack` (100,000,
matching the same reasoning as a companion fix for glyf/gvar's shared-component-subtree issue —
real color-font paint graphs have at most a few hundred nodes). Since every one of the 25
recursive `self.parse_paint(...)` call sites across the format handlers already threads
`recursion_stack: &mut RecursionStack` through, this needed no signature changes beyond
`RecursionStack` itself and its one construction site in `Table::paint`.

---

scadastrangelove:
Closes #221.

Adds a `visits_left: u32` field to `RecursionStack` (initialized to a new
`MAX_PAINT_VISITS: u32 = 100_000` constant in `Table::paint`, its only construction site) and a
`consume_visit()` method, checked in `parse_paint` alongside the existing cycle check:

```rust
// Cycle detected
if recursion_stack.contains(offset) {
    return None;
}

// Total-visit budget exceeded (a DAG can revisit the same offset via different sibling
// branches without ever cycling on the active path -- see RecursionStack::visits_left)
recursion_stack.consume_visit().ok()?;

recursion_stack.push(offset).ok()?;
```

Because `parse_paint` is the single choke point all 25 recursive call sites go through, this
needed no changes to any of them, or to any of the 5 function signatures in the call chain
(`paint` → `paint_impl` → `paint_v1` → `parse_paint` → `parse_paint_impl`).

### Testing
  failure, see #222 for detail).
- depth=18/branching=3 (16–20+s before) now completes in ~3.6ms, `paint()` called 66,658
  times (bounded by the new budget, as expected); depth=20/branching=3 (didn't complete in
  20s before) now completes in ~4ms.

---


Goldziher:
Merging, with two follow-ups I'll apply on top rather than asking you to revise.

**The gap is real and the choke-point analysis is right.** I verified all 26 recursion sites rather than taking the description on trust — every `parse_paint` call in `parse_paint_impl` (formats 1, 4-10, 12-31, and both arms of PaintComposite) plus the PaintColrGlyph path through `paint_impl` → `paint_v1` funnels through `parse_paint`. No site is missed, and it was correct to change nothing else.

The existing `RecursionStack` is a *path* structure, not a visited set: `pop()` runs as soon as the call returns, so `stack[..len]` is the active root-to-leaf path. It bounds height at 64 and forbids a repeat along one path, and says nothing about how many paths exist. With `layers_count` being a `u8`, fan-out reaches 255 per node, so a ~1 KB font gets arbitrarily deep re-walks with no offset ever repeating on any single path. Both existing guards are defeated exactly as you describe.

**No state leak** — `RecursionStack` has exactly one construction site, inside `pub fn paint`, which is the sole entry point and is never called internally. Fresh budget per top-level call, so painting many glyphs in sequence cannot start failing spuriously. That was my main worry and it is clean.

**Follow-up 1 (I'll apply): move the charge above the free bail-outs.** `consume_visit()` currently sits after `Stream::new_at`, the format `read::<u8>`, and the `contains(offset)` cycle check — all three return without charging. An attacker sets 254 of each node's 255 slots to an on-path ancestor (free `contains()` bail) and one slot to the real successor, which makes the effective bound about 100,000 × 256 ≈ 25.5M entries rather than 100,000. Charging first makes the bound exact and matches harfbuzz, which decrements `edge_count` before dispatch.

**Follow-up 2: the constant's comment.** It cites `glyf::MAX_COMPONENT_VISITS` — that symbol didn't exist when you wrote this, but #224 landed a few minutes ago and introduced it at `glyf.rs:437`, so the reference is now valid rather than dangling. I'll still reword it to cite harfbuzz's `HB_MAX_GRAPH_EDGE_COUNT = 2048` as the empirical anchor, since "we picked a number" is weaker than "the production reference ships 2048 and renders Noto Color Emoji." 100,000 stays — being ~48x looser than harfbuzz means zero false-rejection risk.

**One thing outside this diff that I'm filing separately:** `paint_v0` is charged a single visit for a loop of up to `num_layers` = 65,535 iterations. Aim the budgeted nodes at such a record and you still get billions of layer paints. Same vulnerability class, pre-existing, not yours to fix here.

Silent partial paint on exhaustion matches the existing cycle/depth behaviour, so it's consistent rather than a regression. Rebases clean.

Thanks — the fan-out-versus-depth distinction was correctly identified and the single-choke-point claim held up under a full audit.

PR #226, opened 2026-07-22
fix loca.rs to allow fonts with 65536 glyphs.

### LINKED ISSUE #213: Cannot produce outlines for TrueType fonts with 65536 glyphs
For TrueType fonts with the maximum number of glyphs, the code below early returns because the range of u16 is exceeded, leading to `loca` table parsing failing:

https://github.com/harfbuzz/ttf-parser/blob/6e75b3c539ef6bfbacb020aaa3226414ddea7ce5/src/tables/loca.rs#L46-L50

This in turn means that no outlines can be produced.

Having the maximum number of glyphs is not that uncommon in CJK fonts, e.g. Source Han Sans or Noto Serif CJK.

asibahi:
closes #213 (or should, at least).

@RazrFalcon @laurmaedje is this fix sufficient?

(force push was to edit typo in commit)


asibahi:
incidentally the maximum amnount of glyphs is 65535.

Goldziher:
Merging. The impact is worse than the issue describes, and the review turned up why.

There are two defects, and the first masks the second. `u16::try_from(actual_total).ok()?` rejects the table outright for a well-formed 65535-glyph font: 65536 offsets, `try_from` fails, `parse` returns `None`, and `glyf` is then set to `None` wholesale in `collect_tables`. So it is not the last glyph that fails to outline — **every** `outline_glyph` and `glyph_bounding_box` call on that face returns `None`, while `Face::parse` still reports success. Total, silent rendering loss. The separate off-by-one in the `== u16::MAX` branch (which yields 65535 rather than 65536 and loses the final glyph) is real but unreachable for well-formed fonts, because the first defect rejects them first.

**No bounds check is lost**, which was my main concern given the net -16 lines. I accounted for all of them: 3 are type widening, 5 are the replaced `total` block, 1 is `u16::try_from` → `u32::try_from` (check retained, threshold widened), 2 are `read_array16` → `read_array32` (which still validates length through `Stream::read_bytes`), 1 is the `len()` signature. The remaining 4 are the `glyph_id == u16::MAX` guard — and that was never a bounds check. It existed only to stop `glyph_id + 1` overflowing `u16` on the next line, which is exactly how RazrFalcon described it on #213. After widening to `u32` it is dead. The actual bounds check, `if glyph_id + 1 >= self.len()`, is retained verbatim and still rejects GID 65535, since `len()` is provably ≤ 65536. The `min()` clamp against `actual_total` and the spec comment about malformed fonts having fewer glyphs both survive untouched.

Boundaries verified in all four cases: GID 0 reads `[0]`/`[1]`; GID 65534 now reads `[65534]`/`[65535]` where it previously returned `None`; GID 65535 still returns `None`; an empty `loca` still returns `None`. Short and Long are handled symmetrically and the Short ×2 scaling is preserved.

**This is semver-breaking and will be flagged as such.** `loca` is publicly re-exported, the enum variants are public, and both the payload type (`LazyArray16` → `LazyArray32`) and `len() -> u16` → `u32` change. It cannot ship in a patch release. It is going into 0.26.0 under `### Changed` as well as `### Fixed`. In-tree blast radius is zero — nothing in-crate calls `loca::Table::len()`, and the c-api does not expose it.

Follow-ups I'll apply: drop the now-false `// Check for overflow first.` comment, replace the two `as` casts with `u32::from` to match the crate's style, and add `tests/tables/loca.rs` — there is currently no `loca` coverage at all, and `tests/tables/maxp.rs` even carries a `// TODO: what to do when the number of glyphs is 0xFFFF? we're actually checking this in loca` comment which was simply not true.

Thanks for this — it fixes a complete-rendering-failure bug on real pan-CJK fonts.

PR #228, opened 2026-08-02
fix: Ignore the deprecated dotsection operator in CFF charstrings

### LINKED ISSUE #227: CFF glyphs that use the deprecated dotsection operator lose their outlines entirely
## What happens

`outline_glyph` returns `None` for any CFF glyph whose charstring contains the deprecated `dotsection` operator (escape `12 0`). The glyph produces no outline at all, rather than an outline that is merely missing hinting metadata.

## Why

The two-byte operator match in `src/tables/cff/cff1.rs` handles only the flex family (opcodes 34-37) and returns `CFFError::UnsupportedOperator` for everything else, which aborts the entire charstring.

## Impact

Adobe's Type 1 to Type 2 conversion preserves `dotsection` on dot-bearing glyphs, so distiller-produced Type 1C subsets carry it routinely. A PDF rasterizer sitting on ttf-parser silently drops every `i`, `j`, `.`, `:`, `;`, `!`, `?` and `*` on affected pages: "Service refill capacities" renders as "Serv ce ref ll capac t es". The advance width is still consumed by the caller, so the output is gap-filled instead of visibly broken.

## Reproducer

Minimal charstring, in the style of `tests/tables/cff1.rs`:

```rust
test_cs!(dotsection_is_ignored, &[
    CFFInt(10), CFFInt(20), UInt8(operator::MOVE_TO),
    UInt8(12), UInt8(0), // dotsection
    CFFInt(30), CFFInt(40), UInt8(operator::LINE_TO),
    UInt8(operator::ENDCHAR),
], "M 10 20 L 40 60 Z ", rect(10, 20, 40, 60));
```

On 0.25.1 this fails with `UnsupportedOperator`.

## Measured on a real font

Type 1C subset extracted from a production PDF (Century-Light, 11 of 95 glyphs carry `dotsection`), outlined through `Face::outline_glyph`:

| glyph | stock 0.25.1 | with the fix |
|-------|--------------|--------------|
| i | no outline (`UnsupportedOperator`) | OK, 15 segments |
| j | no outline | OK, 17 segments |
| period | no outline | OK, 5 segments |
| colon | no outline | OK, 10 segments |
| l (control) | OK, 10 segments | OK, 10 segments |
| comma (control) | OK, 7 segments | OK, 7 segments |
| n (control) | OK, 21 segments | OK, 21 segments |

Control glyphs are byte-identical before and after, so the change is purely additive.

## Fix

`dotsection` takes no arguments and only brackets the outline sections for the dots in letters such as `i`, `j` and `!` so a hinter can treat them specially. For an outline consumer the correct handling is to clear the argument stack and continue. read-fonts (Fontations, under skrifa) does exactly this in `src/ps/cs.rs`, with a comment naming this very case: "Brackets an outline section for dots in letters such as i, j and !. Purely metadata that a hinter can use."

One-line change in the two-byte operator match, plus the regression test above.

tobocop2:
## Problem

CFF glyphs whose charstrings carry the deprecated `dotsection` operator (escape `12 0`) lose their outlines entirely: the two-byte operator match knows only the flex family and aborts the whole charstring with `UnsupportedOperator`. Adobe's Type 1 to Type 2 conversion preserves `dotsection` on dot-bearing glyphs, so distiller-produced Type 1C subsets drop every `i`, `j`, `.`, `:`, `;`, `!` and `?` through any rasterizer sitting on ttf-parser.

Fixes #227.

## Fix

Treat `dotsection` as a pure no-op in the two-byte match, matching read-fonts ("Nothing to do.") and FreeType. It takes no arguments and only brackets outline sections for hinters.

## Validation

Two new charstring tests: one proves the glyph outlines instead of erroring, one pins the no-op semantics (operands on the stack survive; a `stack.clear()` variant fails it). On the Century-Light subset from the issue, the four dropped glyphs return with correct outlines while control glyphs keep byte-identical segment counts (10, 7, 21), so the change is purely additive. Full suite: 151 passing; clippy warning count unchanged against main.


Goldziher:
Verified against fontations: `read-fonts/src/ps/cs.rs` handles `DotSection` as `// Nothing to do.`, and notably the adjacent `HStem3 | VStem3` arm calls `reset_stack()` while `DotSection` does not — so leaving the operand stack untouched is deliberate upstream semantics, exactly what `dotsection_preserves_stack` pins. Opcode `0` in the two-byte escape namespace matches too.

Rebased onto main (now edition 2024 / MSRV 1.88) and ran the suite on 1.88: clean rebase, 151 tests pass, both new tests green.

Thanks for the thorough writeup and the real-font validation — merging.

PR #229, opened 2026-08-02
Backport the dotsection fix to the 0.25 series for a patch release

tobocop2:
## What this is

The #228 fix (ignore the deprecated `dotsection` operator instead of failing the whole charstring, #227) cherry-picked onto the `v0.25.1` tag, so a 0.25.2 patch release can be tagged directly from this branch.

## Why a backport

Consumers hit the bug through released crates: pdf_oxide, fontdb and anything else outlining Type 1C subsets pin `0.25`. A patch release delivers the fix downstream immediately; the main-branch fix in #228 only ships with the next minor.

## Validation

Identical change and tests to #228. Full suite passes on this branch (151 tests); control glyphs keep byte-identical outlines.

If you would rather cherry-pick onto your own release branch and tag from there, feel free to close this.


tobocop2:
Folding this into #228 to keep one PR. The same commit on the v0.25.1 base lives at tobocop2/ttf-parser:dotsection-0.25.1 if a 0.25.2 patch release is wanted; #228 now points at it.

